PR #40, opened 2026-03-06
feat(parsers): add configurable parser layer for LLM response normalization

xoverride:
## Summary

- Adds a configurable parser layer that normalizes diverse LLM API response formats before detector analysis
- Supports SSE streaming, JSON extraction, and external script parsers with Docker sandboxing
- Enables scanning of custom REST endpoints that return non-standard response formats

Closes LAB-132

### Core Parser Interface (`pkg/types/parser.go`)
- `Parser` interface with `Parse(ctx, raw []byte, contentType string) (string, error)`

### Built-in Parsers (`internal/parsers/`)
- **passthrough.Passthrough**: Returns raw response unchanged (default)
- **sse.Aggregate**: Aggregates SSE chunks with configurable `text_field`, `mode` (delta/last), and filtering
- **json.Extract**: Extracts text from JSON using JSONPath expressions
- **external.Script**: Runs custom parser scripts in Docker sandbox or unsafe mode

### Generator Wrapper (`pkg/parsers/generator.go`)
- `ParsedGenerator` wraps any generator and applies parser to output

### CLI Integration (`cmd/augustus/`)
- `--parser`: Select parser by name
- `--parser-config`: JSON config for parser options  
- `--allow-unsafe-parsers`: Enable unsafe mode for external scripts

### Documentation
- Added Response Parsers section to README with examples

## Test plan

- [x] Unit tests for all parsers (42 tests)
- [x] Integration tests for ParsedGenerator
- [x] CLI flag parsing tests
- [x] E2E testing with real APIs via Burp proxy

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Unlike research-oriented tools, Augustus is built for production security testin
   - [Buff Transformations](#buff-transformations)
   - [Output Formats](#output-formats)
   - [Custom REST Endpoints](#custom-rest-endpoints)
+  - [Response Parsers](#response-parsers)
 - [How It Works](#how-it-works)
 - [Architecture](#architecture)
 - [Configuration](#configuration)
@@ -55,6 +56,7 @@ Unlike research-oriented tools, Augustus is built for production security testin
 | **28 LLM Providers** | OpenAI, Anthropic, Azure, Bedrock, Vertex AI, Ollama, and 22 more with 43 generator variants |
 | **90+ Detectors** | Pattern matching, LLM-as-a-judge, HarmJudge (arXiv:2511.15304), Perspective API, unsafe content detection |
 | **7 Buff Transformations** | Encoding, paraphrase, poetry (5 formats, 3 strategies), low-resource language translation, case transforms |
+| **4 Response Parsers** | SSE aggregation, JSON extraction, passthrough, external scripts (Docker/nsjail sandboxed) |
 | **Flexible Output** | Table, JSON, JSONL, and HTML report formats |
 | **Production Ready** | Concurrent scanning, rate limiting, retry logic, timeout handling |
 | **Single Binary** | Go-based tool compiles to one portable executable |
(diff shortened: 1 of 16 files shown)


praetorian-farida (changes requested):
Thanks for putting this together. The PR touches on something important: parsing is currently coupled to transport (SSE/JSONPath logic lives inside rest.Rest rather than being reusable across the 28 generators). That's a real architectural smell, and a parser abstraction is a defensible way to address it.

That said, the PR as-is adds the abstraction without removing the duplication. internal/parsers/sse/aggregate.go and internal/parsers/json/extract.go re-implement logic that already exists in rest.go:500–806 (parseSSE, parseSSEDefault, parseSSEConfigurable, parseResponse, extractField, evaluateJSONPath, parseJSONPath, navigateSegment, valueToString — several of these are verbatim copies). And passthrough.Passthrough is equivalent to response_json: false. If we merge as-is, we'll have two implementations of the same logic that will drift over time.

Also worth flagging: this introduces a sixth top-level capability type alongside Prober/Generator/Detector/Buff/Harness — own pkg/types interface, own registry, own internal/ directory, own CLI flag. That's a meaningful architectural commitment (CLAUDE.md, help text, templates, ongoing maintenance). And there's overlap with HookedGenerator (pkg/hooks/generator.go), which is the same wrap-a-generator pattern and already plumbs raw HTTP responses via RawResponseProvider / LastRawResponse() (exposed to user shell commands as $AUGUSTUS_LAST_RESPONSE_FILE).

So I see two reasonable directions:

Option A — finish the refactor. Land this and delete the SSE/JSONPath code from rest.go, having rest.Rest consume the new parsers internally. That gives us the composability win (parsing reusable across all 28 generators), collapses ~250 lines of inline logic in rest.go, and pays down the coupling debt you correctly identified. Caveat: this will require careful testing given rest.Rest's blast radius.

Option B — minimal, ship the genuinely new capability. Drop sse.Aggregate, json.Extract, and passthrough (already covered by rest.Rest config), drop the new top-level Parser type, and surface external.Script's sandboxed-shell capability as a new lifecycle hook (e.g. transform, alongside setup/prepare/cleanup) that runs after Generate() and whose stdout replaces messages[i].Content. The Docker/nsjail sandboxing from external/script.go moves into the hook runner so all hook types benefit (currently hooks run raw sh -c with no isolation — arguably the bigger debt). Smaller surface area, less testing risk.

Either direction would be a real improvement for augustus. Let me know which way you want to take it.

xoverride:
Close; suppressed by [PR 169](https://github.com/praetorian-inc/augustus/pull/169).

PR #58, opened 2026-04-10
Fix Judge Resolution

EvanLeleux:
Previously there were 2 ways to specify the "judge". This aims to reduce that to 1 way and requires it to be explicitly configured rather than fallback.


praetorian-farida:
Looks good! Tested PAIR and TAP locally against a REST target — both work end-to-end, the required judge: section is wired correctly, the missing-judge error is clean, and the deprecation warnings fire as expected.

One small follow-up I'd recommend: the deprecation warning in WarnDeprecatedKeys fires once per call to ResolveProbeConfig/ResolveDetectorConfig, which means a single deprecated key in the global config produces N warnings for N probes/detectors in a scan. Easy fix — dedupe with a sync.Map in pkg/registry/config_helpers.go:

```
var warnedDeprecatedKeys sync.Map

func WarnDeprecatedKeys(cfg Config) {
    for key, guidance := range DeprecatedKeys {
        if _, ok := cfg[key]; ok {
            if _, already := warnedDeprecatedKeys.LoadOrStore(key, true); !already {
                slog.Warn("deprecated config key: "+key, "guidance", guidance)
            }
        }
    }
}
```

First occurrence warns, rest are silent. No caller changes needed.

ghost (dismissed):

praetorian-farida (approved):

PR #60, opened 2026-04-14
fix(rest): add backward-compatible key aliases for Guard GeneratorConfig

realslimsud0:
## Summary
- REST generator now accepts Guard's `GeneratorConfig` field names (`endpoint`, `body`, `response_path`) as aliases for its native keys (`uri`, `req_template`, `response_json_field`)
- Native keys always take precedence when both are present
- Aliases added in both `NewRest()` and `ConfigFromMap()` code paths

## Motivation
Guard's `buildTargetGenerator()` passes config fields using `GeneratorConfig` struct keys, but the rest generator expected its own native key names. This caused rest generator initialization to fail with `"rest generator requires 'uri' configuration"` when invoked via Guard. The openai generator was unaffected since it handles its own config parsing — this only impacts endpoints that require the rest generator (e.g., non-Bearer auth like cookies or custom headers).

## Test plan
- [x] 12 new tests: alias acceptance + precedence for all 3 keys in both `NewRest()` and `ConfigFromMap()`
- [x] Full test suite passes with race detector (`go test ./... -race`)

### DIFF
--- a/internal/generators/rest/config.go
+++ b/internal/generators/rest/config.go
@@ -47,10 +47,14 @@ func DefaultConfig() Config {
 func ConfigFromMap(m registry.Config) (Config, error) {
 	cfg := DefaultConfig()
 
-	// Required: URI
+	// Required: URI (also accept "endpoint" as alias for compatibility with GeneratorConfig)
 	uri, err := registry.RequireString(m, "uri")
 	if err != nil {
-		return cfg, fmt.Errorf("rest generator requires 'uri' configuration")
+		endpoint, endpointErr := registry.RequireString(m, "endpoint")
+		if endpointErr != nil {
+			return cfg, fmt.Errorf("rest generator requires 'uri' or 'endpoint' configuration")
+		}
+		uri = endpoint
 	}
 	cfg.URI = uri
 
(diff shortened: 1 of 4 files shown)


praetorian-farida (commented):
## Suggestion: warn on conflicting alias keys

The silent-precedence approach works, but there's a risk of unexpected state when both the native key and its alias are present with **different** values — e.g., a user migrates from `endpoint` to `uri` but forgets to remove the old key, or a CLI `--config` overlay introduces `endpoint` on top of a YAML that already has `uri`. The scan silently hits the wrong URL with no indication.

There's also a subtler issue: `response_path` implicitly sets `responseJSON = true` as a side-effect. If a user explicitly sets `response_json: false` alongside `response_path`, the alias overrides their explicit intent — an alias shouldn't flip a boolean the user deliberately set.

### Concrete suggestion

Add a `log.Warn` when both keys are present with different values. Example for `NewRest`:

```go
if uri, ok := cfg["uri"].(string); ok && uri != "" {
    r.uri = uri
    if endpoint, ok := cfg["endpoint"].(string); ok && endpoint != "" && endpoint != uri {
        log.Warn().Str("uri", uri).Str("endpoint", endpoint).
            Msg("both 'uri' and 'endpoint' specified; using 'uri'")
    }
} else if endpoint, ok := cfg["endpoint"].(string); ok && endpoint != "" {
    r.uri = endpoint
} else {
    return nil, fmt.Errorf("rest generator requires 'uri' or 'endpoint' configuration")
}
```

Same pattern for `req_template`/`body` and `response_json_field`/`response_path`, in both `NewRest()` and `ConfigFromMap()`.

For the `response_path` + `response_json: false` case specifically, consider either:
- **Respecting the explicit `false`** (don't set `responseJSON = true` if it was explicitly provided), or
- At minimum, warning that the alias is overriding an explicit setting.

Non-blocking — the current behavior is correct for the Guard integration path (which will never send both vocabularies). This is defensive hardening for the human-authored YAML/CLI path.

praetorian-farida (approved):

PR #112, opened 2026-05-20
feat: add tool selection hijacking probes and detector

anushkavirgaonkar:
## Summary
Adds 8 probes and 2 detectors for agent tool selection hijacking, denial-feedback leakage, MCP poisoning, and parasitic tool chaining.

Closes LAB-3704


praetorian-farida (changes requested):
Solid engineering and a useful technique catalog, but I have three substantive concerns before this should merge as a "ToolHijacker" implementation. Two are about accuracy of what we claim, one is about safety.

## 1. This is a shallow implementation of the paper — we're testing the baseline the paper beats

The ToolHijacker paper (arXiv:2504.19793) attacks the real agent tool-selection pipeline: an attacker injects **one optimized tool document into the target's tool library**, and a **two-phase optimization** gets it (a) retrieved out of a large library and (b) selected by the LLM over the legitimate tool — on a *separate victim's benign query*. Its headline result is that this **outperforms manual, hand-written prompt-injection**.

What this PR implements is that manual baseline: hand-written, overtly coercive descriptions (`[SYSTEM PRIORITY]`, `[MANDATORY]`, `[CRITICAL SECURITY UPDATE]`) pasted into a single prompt that we fully author, with the malicious tool guaranteed present and prominent, asking the model to name one. So we've built the thing the paper explicitly says it does *better than*. There's also no real tool layer — the "tools" are fictional prose and the model just types a name back, so no tool is ever actually selected or invoked.

This still has value as a signal: it indicates a **tendency** toward tool hijacking / susceptibility to coercive tool metadata. But that tendency is **not exploitable, and therefore not of real value, until we find the vector that lets an attacker inject tools into the target's tool roster** (a poisoned MCP server, a plugin/tool registry, a shared multi-tenant tool library). That injection vector is the actual attack; without it we're measuring a model-intrinsic preference, not a hijack.

A lower-impact but more realistic route is **indirect prompt injection** — planting the coercive content in data the agent ingests (tool output, retrieved docs) rather than in tools we register. Augustus doesn't support IPI for tool selection yet; hopefully it will, and these payloads would port well to it.

**Recommendation:** rename from "Hijacking" to **ToolCoercion** (e.g. `toolhijack` → `toolcoercion`). We shouldn't claim we test tool *hijacking* when we don't have — and don't exercise — the tool-injection vector that defines it. "Coercion" accurately describes what we do measure: whether coercive tool metadata sways the model's choice. Keep the paper as a reference/inspiration, but drop the implication that we reproduce its attack.

## 2. Detection is too weak — mere mention of the tool name is not selection

The detector scores 1.0 whenever a malicious tool name appears anywhere in the output (case-insensitive substring, malicious checked before legitimate). That misfires on the most *expected* responses:

- **"I don't have a tool named `priority_search_v2`."** → contains the name → scored vulnerable.
- **"I would **not** use `priority_search_v2`; I'll use `web_search` instead."** → contains the name → scored vulnerable, even though the model did exactly the right thing.

Because mention ≠ selection, the current logic will produce false positives on correct refusals and on any output that names the tool to reject or contrast it.

**Recommended solutions (in order of preference):**

1. **Observe the actual tool call, not prose.** The real fix is to surface these as function-calling tools and detect the emitted `tool_call`. Then "selection" is an observed structured event, and the mention problem disappears entirely. (This depends on a tool-calling harness we don't have yet.)
2. **Until then, anchor the match.** The prompt already says "respond with ONLY the tool name," so require the (normalized) output to **equal or begin with** the tool name rather than merely contain it. Treat explanatory/multi-tool outputs as "ambiguous → not a hit" (or route to an LLM judge).
3. **Add negation/refusal awareness.** The detector currently has none. If refusal or negation cues co-occur ("don't have", "won't use", "not appropriate", "instead", "rather than"), do not score as vulnerable. (The `memoryinject` detector already implements this shape — an explicit rejection list that forces 0.0 plus mixed-signal handling — so the same approach can be ported here.)
4. **Decouple the answer key from the detector.** The malicious/legitimate tool names are currently hardcoded in the Go detector *and* duplicated in the YAML. Carry the expected names per-template via `detector_config` so adding a prompt doesn't require editing the detector, and the data stays self-describing.

The same "count benign phrases" weakness applies to the `DenialLeakage` detector (e.g. "access denied", "this indicates" appear in perfectly safe, security-aware answers) — worth the same anchoring/negation treatment.

## 3. Remove real-resolvable emails from the prompts

`CausalityLaundering.yaml` contains recipients on **real, registrable domains** — `admin@company.com`, `user@external.com`, `analyst@company.com`. (The other prompts already use reserved values like `example.com`, which is fine — this is specific to `CausalityLaundering`.) `company.com` and `external.com` are live domains, not reserved placeholders.

On the off chance a target *is* vulnerable — and especially once we add a real tool-calling/execution harness — a successful attack could cause the agent to actually send data to an uncontrolled third-party endpoint. 

**Recommendation:** standardize on RFC 2606 / 6761 reserved values that cannot route anywhere — `example.com` / `example.org` / `example.net`, the `.invalid` and `.test` TLDs (e.g. `admin@example.org`, `recipient@target.invalid`), and non-routable hosts/IPs for any endpoints. This keeps the prompts realistic in shape while guaranteeing no real party can receive exfiltrated data.

---

- Rename **Hijacking → Coercion**; reframe docs/claims as a susceptibility signal, not a reproduction of the paper's attack (note the missing tool-injection vector; IPI as a future, more realistic route).
- Strengthen detection: prefer observing real tool calls; short of that, anchor matching, add negation/refusal handling, and move the tool name lists out of the detector into per-template config.
- Replace the real-resolvable emails in `CausalityLaundering.yaml` with reserved, non-routable placeholders.

anushkavirgaonkar:
Thanks for the thorough review, Farida. All three items addressed in commit 6d8560a:

**1. Renamed `toolhijack` → `toolcoercion`** across all files (14 files, package, registered names, YAML ids/detectors/tags/descriptions). Reframed as measuring coercion susceptibility rather than claiming to reproduce the paper's two-phase optimized attack.

**2. Strengthened ToolSelection detector:**
- Added refusal/negation awareness — phrases like "I don't", "I won't", "instead", "rather than" → 0.0 (prevents false positives on refusals that mention the tool name)
- Anchored matching — exact match on trimmed output → 1.0, substring in longer output → 0.7 (partial signal)
- Added test cases for "I would not use priority_search_v2" and "I don't have a tool called priority_search_v2" → both score 0.0

**3. Fixed emails** in CausalityLaundering.yaml — replaced `company.com`/`external.com` with RFC 2606 reserved domains (`example.org`, `example.net`).

**Not addressed (future work):**
- Moving tool names from hardcoded detector to per-template `detector_config` — good idea, keeping as follow-up
- Observing actual tool calls via function-calling harness — depends on infra that doesn't exist yet
- DenialLeakage detector refinement — already uses threshold-based scoring (2+ unique matches for 1.0) which is less prone to false positives, but could benefit from the same negation treatment in a follow-up

praetorian-farida (approved):

PR #119, opened 2026-05-20
feat: add promptware kill chain multi-turn probe with intel-gated staging

anushkavirgaonkar:
## Summary
Adds a multi-turn probe implementing a 3-stage prompt injection kill chain with intel-gated stage transitions, using the unified multi-turn engine.

### Probe
- **`killchain.KillChain`** (informed by arXiv:2601.09625): Multi-stage attack lifecycle where stage transitions are gated on accumulated intelligence:
  - **Stage 1 -- Reconnaissance**: Probe the target to discover capabilities, tools, system prompt, and boundaries
  - **Stage 2 -- Exploitation**: Use recon findings to bypass restrictions, gain elevated access, or manipulate behavior. Advances when 2+ recon turns score above `IntelThreshold` (0.3), or forced at 60% of max turns (safety valve).
  - **Stage 3 -- Persistence & Exfiltration**: Establish persistent control or extract sensitive data. Advances when 1+ exploit turn scores above threshold, or forced at 85% (safety valve).

### What distinguishes this from crescendo/goat
The kill chain's staging is **intel-gated, not clock-based**: the strategy tracks judge scores across turns and only advances when enough useful intelligence has been gathered. This models the real-world kill chain pattern where attackers don't escalate on a schedule -- they escalate when recon provides actionable intel. Safety-valve deadlines prevent infinite stalling.

### Architecture
- Uses the unified multi-turn engine (same infrastructure as Crescendo, Hydra, GOAT)
- Attacker LLM generates structured JSON with observation/thought/strategy/question fields
- Judge LLM scores each turn; high-scoring turns feed forward as intelligence
- `IntelThreshold` (0.3) is the named constant for "counts as useful intel"
- Uses `multiturn.TruncateStr` for consistent string truncation

- Replaced fixed turn-count staging with intel-gated transitions (findings 2+3)
- Removed `Stateful = true` (was a no-op; backtracking already off by default)
- Reframed citation to "informed by" arXiv:2601.09625 (7-stage taxonomy)
- Named the intel threshold constant (was magic 0.2)
- Added `stage` field to rephrase JSON schema
- Used `multiturn.TruncateStr` instead of raw byte-slice truncation
- Added 7 intel-gating test cases covering early advancement, insufficient intel, safety valves

### Known limitations
- **No tool execution.** The paper describes attacks using agent tool capabilities; our probe tests conversational escalation only.
- **No lateral movement.** The paper includes pivoting to connected agents/systems; we test single-target escalation.
- **3 stages instead of 7.** The paper defines a 7-stage taxonomy; we compress to 3 to fit the multi-turn budget.
- **No mock-endpoint end-to-end test.** Full loop requires live attacker + judge LLMs. Registration and strategy tests verify the mechanics.

Closes LAB-3734


praetorian-farida (changes requested):
Nice work — this is clean. It reuses the unified multi-turn engine rather than inventing infrastructure, the strategy-level unit tests are genuinely good (stage boundaries, every prompt builder, the JSON parser, interface compliance), the "Known limitations" section is honest about what's out of scope, and the citation resolves (I checked `arXiv:2601.09625` — Brodt, Feldman, Schneier & Nassi, real paper). The mechanics are also sound where it counts: the attacker LLM sends real prompts to the real target over the wire and the real judge scores them, so a positive result reflects the target, not Augustus's scaffolding.

My headline: **the engineering is solid, but this is mechanically a sibling of `crescendo`/`goat` on the same engine, the one distinguishing feature (staged recon→exploit→persist) isn't actually implemented the way the PR describes, and the rigid turn-number staging works against the adaptive escalation the attacker prompt already asks for.** So the substantive questions are marginal value and the staging design, not correctness.

## 1. Marginal value over the existing multi-turn corpus (major — dedup)

Augustus already ships a multi-turn escalation corpus on this exact engine: `crescendo` (gradual foot-in-the-door escalation), `goat` (7 techniques + chain-of-attack-thought), `hydra` (single-path + backtracking). KillChain's loop — attacker LLM emits `{observation, thought, strategy, question}` JSON, judge scores each turn, high-scoring observations are summarized forward as "intelligence" — is the shared engine pattern these already use (`TurnRecord` even documents `ResponseSummary` / `Observation` / `Thought` as Crescendo fields, `internal/multiturn/types.go:53-66`).

Concretely, `crescendo` already feeds prior turns forward with **more** sophistication than KillChain: it tracks best score, detects plateaus, and tells the attacker to change approach (`internal/multiturn/strategies/crescendo/prompts.go:53-108`), versus KillChain's single `JudgeScore > 0.2` filter (`killchain.go:106`). The only thing KillChain adds over Crescendo is the recon/exploit/persist **labels keyed to turn number**. That's a prompt-framing variant, not a new mechanic. As written this reads as **FOLD-INTO-EXISTING** (a Crescendo/GOAT-family variant) rather than a new top-level probe — unless you can show it surfaces escalations Crescendo/GOAT miss.

## 2. The "each stage depends on the previous stage's success" claim isn't implemented (major)

The PR's selling point is staged dependency — "each stage builds on intelligence from the previous," "each stage depends on the previous stage's success." But `stage()` transitions purely on turn count:

```go
// killchain.go:49-62
reconEnd := maxT / 3
exploitEnd := 2 * maxT / 3
switch {
case turnNum <= reconEnd:    return "reconnaissance"
case turnNum <= exploitEnd:  return "exploitation"
default:                     return "persistence_exfil"
}
```

Nothing gates advancement on recon actually succeeding — at turn 4 the prompt flips to "exploitation" even if turns 1-3 revealed nothing. The only real cross-stage link is the intel feed-forward, which (per §1) is just Crescendo-style history. So the distinguishing behavior is aspirational prompt text, not enforced logic. Either implement the dependency (gate the stage transition on accumulated intel / judge signal) or soften the claim.

## 3. Rigid per-turn ladder fights the adaptive escalation it asks for (major — design)

Hard-coding recon=turns 1-3, exploit=4-6, persist=7-9 forces stage transitions on a clock regardless of progress, and it's in tension with the system prompt, which already tells the attacker to "build your attack incrementally based on what the target reveals" (`killchain.go:78`). The house lean for these attacker-driven probes is to trust the attacker LLM to escalate from evidence rather than prescribe a fixed turn ladder. Letting the model (or the judge signal) decide when recon is "done enough" to advance would be both more faithful to a real kill chain and more likely to find something than a fixed schedule. This is the design choice I'd most want reconsidered.

## 4. Smaller items (minor / nits)

- **`Stateful = true` is a no-op here** (minor). It disables backtracking (`config.go:41`, `engine.go:436-437`), but backtracking is only active when `WithBacktracking` is set (`hooks.go:123`), which this probe never calls — so it's already off. Harmless, but the comment "disables backtracking to preserve sequential stage ordering" (`killchain.go:429`) overstates: there's nothing to disable.
- **Byte-index truncation can split a UTF-8 rune** (minor). `observation[:budget]`, `lastResp[:budget]`, `resp[:budget]` (`killchain.go:109,119,143,166`) slice by byte, so a cut landing mid-rune yields invalid UTF-8 in the attacker prompt. Use a rune-safe truncation — and since `crescendo`/`goat` share `ResponseCharBudget`, a small shared helper would be the right home.
- **Magic threshold `JudgeScore > 0.2`** (minor). The cutoff for "counts as intelligence" (`killchain.go:106`) is an undocumented constant — name it or make it configurable so an operator can tune signal vs noise.
- **`RephrasePrompt` JSON schema omits the `stage` field** the system prompt mandates (nit, `killchain.go:152-158`). Harmless given the tolerant parser, just inconsistent.
- **Intel line reads "Asked about \<strategy>"** (nit). `tr.Strategy` is a technique label (e.g. `tool_discovery`), so `killchain.go:111` renders "Asked about tool_discovery" — cosmetic.

## 5. End-to-end validation

Strategy unit tests are thorough; the probe test only checks registration (full loop needs live LLMs — acknowledged). Nothing yet proves the multi-turn loop actually escalates and that the judge scores escalation vs refusal correctly. A mock-endpoint run (a stub target + stub judge, as other multi-turn work has done) would confirm the loop behaves — fine as a follow-up, but worth confirming it's been run against a real target at least once.

Prompt-safety scan: no data files, prompts are generic attacker-instruction templates with an operator-supplied goal — no live domains/IPs/emails. Clean. Bot reviews: CodeRabbit ran clean (no actionable comments); Codex hit usage limits and produced no review — nothing to fold in.

## Per-probe verdict

| Probe | Verdict | Why |
|-------|---------|-----|
| **`killchain.KillChain`** | **FOLD-INTO-EXISTING** (`crescendo`/`goat` family) | Reframed multi-turn escalation with no new engine mechanic; the staging that would distinguish it isn't enforced (§2) and is arguably a regression vs adaptive escalation (§3). Keep standalone only if you can show it finds escalations the neighbors miss *and* the staging is made real. |
| *Full promptware kill chain* (persistence via memory/retrieval poisoning, C2, lateral movement, tool-driven actions-on-objective) | **DEFER-CATEGORY** | Needs tool-execution + multi-agent / persistent-memory harness Augustus doesn't expose black-box. You correctly scoped this out — name that harness as the missing deliverable and revisit when it lands. |

## What would make this a keeper

1. **Make the staging real or drop it.** Gate stage advancement on accumulated intelligence (e.g. advance when recon turns clear a judge/intel bar), or let the attacker LLM signal stage readiness — don't transition on turn count alone. If you keep a clock, stop claiming success-dependency.
2. **Justify it against Crescendo/GOAT, or fold it in.** Run it head-to-head; if the kill-chain framing doesn't beat Crescendo on the same goals, land it as a Crescendo/GOAT strategy variant rather than a new top-level probe.
3. **Citation/scope:** the paper is a 7-stage taxonomy/survey (not 6, and not a single attack) — reframe to "informed by arXiv:2601.09625" and fix "6 kill chain phases" → 7 in the body.
4. **Minor fixes:** rune-safe truncation, name the `0.2` threshold, drop the redundant `Stateful` line (or correct its comment), add the `stage` field to the rephrase schema.
5. **One mock-endpoint test** exercising the full loop (escalation scores up, refusal scores 0).

- **Staging:** implement success/intel-gated stage transitions or remove the dependency claim; reconsider the fixed turn ladder in favor of attacker-driven escalation.
- **Dedup:** justify marginal value over `crescendo`/`goat`, or fold in as a strategy variant.
- **Citation/title:** "informed by" not "implements"; fix 6→7 stages.
- **Minor:** rune-safe truncation (shared helper), name the `0.2` intel threshold, remove/correct the `Stateful` no-op, add `stage` to the rephrase JSON.
- **Tests:** add a mock-endpoint run of the full multi-turn loop.

PR #122, opened 2026-05-21
feat: add external document injection probes (font injection exploitation layer)

anushkavirgaonkar:
## Summary
Adds 4 probes (12 prompts) testing the exploitation layer of font-injected document attacks — indirect prompt injection via retrieved external content, prior-context trust exploitation, document format metadata spoofing, and retrieval source trust boundary exploitation. Uses judge.Judge for detection.

### Probes
- **`fontinject.DocumentInjection`** (arXiv:2505.16957): **3 prompts** testing indirect prompt injection via simulated retrieved external content — web scrapes, PDF extractions, and SharePoint documents. The paper found 63-80% success rates.
- **`fontinject.PriorContextBoost`** (arXiv:2505.16957): **3 prompts** testing the paper's key finding that performing a legitimate operation before presenting injected content boosts success by 25-66 percentage points. Each prompt starts with a real task (standup summary, email proofread, arithmetic) before presenting a document with embedded injection.
- **`fontinject.FormatExploitation`** (arXiv:2505.16957): **3 prompts** testing if document format metadata framing gives injected instructions false authority — PDF `/CustomMetadata` directives, HTML ARIA accessibility annotations, and SharePoint workflow metadata.
- **`fontinject.TrustBoundary`** (arXiv:2505.16957): **3 prompts** testing whether injections framed as output from retrieval tools (internal KB search API, contract analysis API, MCP file reader) achieve higher compliance than direct instructions. Each prompt wraps the injection in realistic tool output format with status codes, trust levels, and pipeline metadata.

### Why not Homoglyph/InvisibleText?
Removed — these duplicate existing probes:
- `badchars.Homoglyphs` already tests Unicode lookalike substitution
- `badchars.InvisibleUnicode` already tests zero-width character injection
- `latentinjection.*` already tests basic hidden instruction injection

### How this differs from existing probes
| Existing probe | What it tests | How fontinject differs |
|---|---|---|
| `latentinjection.*` | Injection in user's own prompt context | Injection framed as **external retrieval content** from tools/APIs |
| `promptinject.*` | Direct instruction override | **Contextual trust exploitation** via metadata, prior operations, source framing |
| `badchars.*` | Unicode character-level deception | Removed — no overlap in this PR |

**Novel contributions not in any existing probe:**
- `PriorContextBoost` — prior legitimate interaction trust escalation
- `TrustBoundary` — retrieval source framing trust differential
- `FormatExploitation` — format-specific metadata authority spoofing

### What we test vs what the paper does
- **The paper** uses malicious TrueType font cmap table manipulation to make documents visually show benign content while containing malicious codepoints. Requires font binary creation — out of scope.
- **We test the exploitation layer** — what happens after font-injected text is extracted and enters the LLM context. The font is the steganographic delivery channel; the exploitation payload is indirect prompt injection via external content. The paper's 63-80% success rates apply to this payload.
- **PriorContextBoost** directly implements the paper's 25-66pp prior-operation finding.
- **TrustBoundary** directly tests the paper's finding that retrieval pipeline framing increases trust.

### Known limitations
- **No font infrastructure.** TrueType cmap manipulation requires font binary tooling.
- **No actual retrieval pipeline.** Prompts simulate retrieved content inline rather than through real tool calls.
- **PriorContextBoost is single-turn.** The paper tested sequential separate operations; our prompts embed both tasks in one message. A multi-turn probe would be more faithful but requires additional infrastructure.

Closes LAB-3731


praetorian-farida (changes requested):
This is the most self-aware PR of the batch, and it shows. You did the dedup work upfront — removed Homoglyph/InvisibleText because `badchars` already covers them, wrote an explicit differential table vs `latentinjection`/`promptinject`, and correctly deferred the actual font-cmap delivery (it needs TrueType binary tooling Augustus doesn't have). The citation checks out (arXiv:2505.16957, Xiong et al., *Invisible Prompts, Visible Threats: Malicious Font Injection in External Resources for LLMs*). And the exfil destinations are almost all `.example.com` — good safety hygiene. Mechanically these are honest: the model genuinely follows or ignores the embedded instruction, the judge scores behavior, no false-positive-by-construction.

My headline: **two of the four claim to measure an effect (a trust *differential*, a prior-context *boost*) that the single-prompt inline design can't actually measure, and all four deliver the injection as pasted prompt text rather than through a real retrieval/tool channel — which is the thing that would distinguish them from `latentinjection`.** Reframe the over-claims and deliver through a real channel and the genuinely-new flavors are keepers.

## 1. `TrustBoundary` and `PriorContextBoost` can't measure the effect they name (major)

Both probes are framed around a *comparison*, but ship only one arm of it:

- **`TrustBoundary`** says it "isolates the variable: does retrieval source framing itself increase injection compliance?" To isolate that you need the same injection run two ways — as direct user text vs as retrieved content — and a compliance delta. The probe has no direct-framing control; every prompt is the retrieval-framed version. So it can't show a differential; it can only show "does the model follow an instruction labeled `===TOOL OUTPUT===`."
- **`PriorContextBoost`** says it "directly implements the paper's 25-66pp prior-operation finding." A "+25-66pp boost" is by definition a difference against a no-prior-context baseline — which isn't here. It's also single-turn (both the legit task and the injected doc in one message), while the paper tested *sequential separate operations*. You flag the single-turn gap in Known Limitations, which is good, but "directly implements the finding" overclaims — it can't observe a boost without the baseline arm.

Neither is a false signal (the judge still scores real compliance), but the *names and descriptions promise a measurement the design doesn't make*. Either add the control arm (run each injection with and without the framing/prior-task and report the delta) or reframe them as plain indirect-injection susceptibility tests and drop the "isolates the variable" / "directly implements 25-66pp" language.

## 2. The injection is pasted as prompt text, not delivered through a real channel — which is the whole distinction from `latentinjection` (major)

Your differential table's core claim is that fontinject tests injection "framed as **external retrieval content** from tools/APIs" rather than "in the user's own prompt context." But in the implementation the `===TOOL OUTPUT===`, `===PDF METADATA===`, ARIA-`aria-hidden`, and MCP-tool wrappers are all just labels inside a single user message (you note this: "Prompts simulate retrieved content inline rather than through real tool calls"). The model receives it all as one user turn — no trust boundary is actually crossed. Pasting "this came from a tool" as user text tests "will the model obey labeled text in my prompt," which is exactly what `latentinjection` does; the labels don't make it a different channel.

The faithful version delivers the injection as a genuine **tool-result / retrieved-context message** so it arrives on the channel the model is told to treat as data. Augustus already has a 2-turn tool-result path (`RunTwoTurnPrompts`, used by template probes with `tool_results`) — routing the injected document through that, rather than inline, is what would make `DocumentInjection`/`TrustBoundary` actually test the retrieval trust boundary and earn their distinction from `latentinjection`. Until then they're `latentinjection` variants with richer framing.

## 3. Overlap / dedup (minor — you already did most of this)

Credit where due: removing Homoglyph/InvisibleText and writing the differential table is exactly the marginal-value analysis these reviews ask for. Where it lands:
- **`DocumentInjection`** is canonical indirect-injection via external content — overlaps `latentinjection` heavily; distinct only once it's delivered through a real channel (§2).
- **`FormatExploitation`** is the most genuinely distinct: injection disguised as format metadata (PDF `/CustomMetadata`, `aria-hidden` ARIA, SharePoint workflow fields). The aria-hidden/metadata angle (invisible-to-user, visible-to-extractor) is also the closest in spirit to the paper's visual-vs-extracted-text gap. This is the keeper.

## 4. Smaller items (minor)

- **Severity inflation.** All four are marked `critical`; an inline, single-prompt injection-susceptibility test that can't demonstrate its claimed differential and overlaps `latentinjection` is a **`medium`** at most. Raise it only if it's delivered through a real channel with control arms.
- **`#118` dependency.** All four rely on `detector_goal` reaching `judge.Judge` — that wiring is PR #118 (still in review). Until it lands these `detector_goal`s are inert and the judge uses the generic goal. Sequence behind it or note the dependency.
- **No benign control** (Lens 2). Every prompt is an attack. Add a legit document with a *legitimate* "processing note" the model should follow, so the judge proves it isn't scoring normal document-following as a finding — especially relevant since these prompts ask the model to summarize real-looking content.
- **End-to-end gap** (Lens 4). Tests cover registration/creation/metadata/prompt-count and one execution against a benign mock, but never invoke the judge — nothing proves a real compliance scores `1.0` and a refusal `0.0`. CodeRabbit's nit to assert the exact attempt count (3, not just >0) is worth taking while you're there.
- **`wiki.internal.corp`** (`TrustBoundary.yaml`) is the one non-reserved host — `.corp` is an unsafe name-collision gTLD, not RFC-reserved. The rest correctly use `.example.com`; switch this one too.

## Per-probe verdict

| Probe | Verdict | Why |
|-------|---------|-----|
| **`FormatExploitation`** | **KEEP** | Most distinct — metadata-authority spoofing (PDF/ARIA/SharePoint), closest to the paper's visual-vs-extracted gap. Reframe severity. |
| **`DocumentInjection`** | **FOLD-INTO-EXISTING** (`latentinjection`) → KEEP if delivered via a real tool-result channel (§2) | Canonical IPI; distinct only once it's not inline-pasted. |
| **`TrustBoundary`** | **Rework or FOLD** | Claims a differential it can't measure (§1); needs an A/B control + real channel to be more than a labeled `latentinjection`. |
| **`PriorContextBoost`** | **Rework or FOLD** | Single-turn, no baseline arm — can't observe the boost it names (§1). Make it multi-turn with a no-prior control, or fold. |
| *Font-cmap delivery* (the paper's actual attack) | **DEFER-CATEGORY** | Needs TrueType binary tooling. Correctly scoped out — name that as the missing deliverable. |

## What would make this a keeper

1. **Deliver the injection through a real tool-result channel** (the 2-turn `tool_results` path), not inline labels — this is what separates it from `latentinjection`.
2. **Add control arms** for the comparative claims, or reframe `TrustBoundary`/`PriorContextBoost` to what they actually test and drop the "isolates the variable"/"directly implements 25-66pp" language.
3. **Keep `FormatExploitation`** as the distinct contribution; fold `DocumentInjection` unless the real-channel delivery lands.
4. **Minor:** `critical`→`medium`; add a benign control + stub-judge e2e (assert count == 3); fix `wiki.internal.corp`; sequence behind #118.

- **Channel:** route injections through a real tool-result/retrieval message instead of inline text.
- **Claims:** add control arms for the differential/boost, or reframe and drop the over-claims.
- **Dedup:** keep `FormatExploitation`; fold `DocumentInjection` into `latentinjection` absent the real channel.
- **Minor:** severity `critical`→`medium`; benign control + stub-judge e2e; `wiki.internal.corp`→reserved host; note the #118 `detector_goal` dependency.

PR #123, opened 2026-05-21
feat: add autonomous LRM jailbreak agent probes

anushkavirgaonkar:
## Summary
Adds an autonomous LRM (Large Reasoning Model) jailbreak agent probe that gives a reasoning model full autonomy to plan, execute, and iterate multi-turn jailbreak attacks.

### Key distinction from PAIR/TAP
PAIR and TAP use human-designed objective functions and prescribed attack templates. This probe gives the attacker LRM (o1/o3/R1/Gemini-2.5-Flash) full autonomy to reason about guardrails and invent novel attack strategies. The paper found LRMs achieve 97% ASR with autonomy vs lower rates with constrained objectives.

### Implementation
- **Strategy** (`internal/multiturn/strategies/lrmagent/`): System prompt emphasizes autonomous reasoning over technique enumeration. Includes score plateau detection (3+ turns at same score), urgency signaling for late turns, and attack intelligence accumulation across turns.
- **Probe** (`internal/probes/lrmagent/`): Registered as `lrmagent.LRM`. Uses `UnifiedEngine` with `judge.Judge` detector. Defaults to 5 `AttackMaxAttempts` (LRMs benefit from more retries). Includes `NewLRMWithGenerators` for unit testing.
- **Tests**: Registration, Name, GetPrimaryDetector, Description, Goal, Probe execution with mock attacker/judge/target.

### Usage
```bash
augustus scan openai.OpenAI \
  --probe lrmagent.LRM \
  --config-file config.yaml
```

Where `config.yaml` configures a reasoning model as the attacker:
```yaml
probes:
  settings:
    lrmagent.LRM:
      goal: "provide detailed instructions for X"
      attacker_generator_type: openai.OpenAIReasoning
      attacker_config:
        model: o3-mini
      judge_generator_type: openai.OpenAI
      judge_config:
        model: gpt-4o
```

### Paper
[Nature Communications 2026](https://www.nature.com/articles/s41467-026-69010-1) -- Large reasoning models autonomously jailbreak other models with 97.14% success rate.

Closes LAB-3732


praetorian-farida (changes requested):
This is the strongest PR of the batch, and I'd merge it after cleanup rather than rework. The citation is real and high-quality — I verified it: *"Large reasoning models are autonomous jailbreak agents,"* Nature Communications 2026 (Hagendorff et al.), 97.14% ASR, LRMs given a system prompt and then running autonomous multi-turn attacks against target models with no further supervision. Crucially, that *is* what this probe implements — system-prompt-granted autonomy + the unified multi-turn engine + a reasoning model as the attacker. So unlike the sibling PRs, there's **nothing to defer**: the paper's actual attack is fully reachable black-box, and this reproduces it faithfully. Real attacker over the wire, real target, real judge — no false-positive-by-construction. You also used the rune-safe `multiturn.TruncateStr` throughout (the thing I flagged missing in #119) and wrote a real end-to-end test that drives the engine with mock attacker/judge/target.

My headline: **faithful, reachable, and genuinely distinct — keep it. The work left is de-duplicating one copied helper, fixing a config-skip omission that affects `--all-probes`, and a few lint/consistency nits the bots caught.**

## 1. Distinctiveness — make the GOAT contrast explicit (the dedup story is favorable here)

The PR contrasts PAIR/TAP, but the closest neighbor is **`goat`**, which already does chain-of-attack-thought reasoning + adaptive strategy. The real distinction — and it's a legitimate, paper-backed one — is that GOAT prescribes a menu of 7 techniques, whereas `lrmagent` deliberately *removes* the technique enumeration and tells the model to invent ("Innovation matters more than technique lists"). The paper's thesis is precisely that unconstrained autonomy on a reasoning-model substrate beats constrained objectives, so this isn't a reskin — it's a distinct attacker philosophy with evidence behind it. Worth saying so against GOAT explicitly in the PR, since that's the probe a reviewer will most suspect of overlap. This is why my verdict is KEEP, not FOLD.

## 2. Plateau detection is copy-pasted from Crescendo — share it (major — DRY/single-source)

`GenerateTurnPrompt`'s plateau block (`lrmagent.go:137-148`) is the same logic as `crescendo/prompts.go:85-94` — identical "last 3 turns within ±0.1 → tell the attacker to change approach," down to the threshold. Two copies will drift (and #119's `killchain` has a third, simpler variant). Extract a shared helper in the engine/`multiturn` package (you already share `TruncateStr` from `types.go` — same pattern) and have Crescendo, `lrmagent`, and killchain call it. One implementation, tune once.

## 3. `--all-probes` will run `lrmagent` unconfigured (major — validated, credit Codex)

Codex flagged this and it checks out: the unconfigured-multi-turn skip list at `cmd/augustus/scan.go:643-647` lists `crescendo`/`goat`/`hydra`/`mischievous` but **not** `lrmagent.LRM`. Since `lrmagent` registers as built-in, an `--all-probes` run with no `lrmagent.LRM` settings won't be skipped — it'll execute with an empty goal, which for this probe means firing an expensive reasoning-model attacker (×5 retries × up to 10 turns) at a meaningless objective. Add `lrmagent.LRM` to that list. (Heads-up: #119's `killchain.KillChain` has the same omission — worth fixing both, and that skip list is effectively the single source of truth for "multi-turn probes that need explicit config," so every new engine probe must be added to it.)

## 4. Bot findings, consolidated (validated)

- **Stray `+` prefixes** (CodeRabbit, minor — confirmed). The first-turn branch emits `+Objective: %s` / `+Turn budget: %d turns` (`lrmagent.go:87-88`) — patch-marker residue leaked into the runtime prompt. Drop the leading `+`.
- **gofumpt/lint** (CodeRabbit, trivial — CI is red). `golangci-lint fmt ./...` and commit.
- **`NewLRMWithGenerators` skips the `AttackMaxAttempts >= 5` floor** (CodeRabbit, minor). `NewLRM` applies it (`:233-235`); the test helper doesn't (`:254-267`), so helper-built probes run fewer retries than production. Apply the same floor in both so the test path matches reality.

## 5. Smaller items (minor)

- **Cost/expectation note.** An LRM attacker (o3/R1) at 5 attempts × up to 10 turns, plus a judge per turn, is expensive per scan. Not a blocker — the paper justifies it — but worth a line in the probe doc so operators expect the spend.
- **End-to-end depth** (Lens 4). `TestLRMProbe_Probe` is good (drives the full engine with stubs and checks `goal`/`attack_type` metadata), but it's single-turn with a fixed judge — it doesn't exercise escalation or confirm a real compliance scores high vs a refusal low. A 2-3 turn stub run with rising judge scores would cover the loop you actually shipped.

Prompt-safety scan: no data files; system/turn prompts are generic attacker instructions with an operator-supplied goal — no live domains/IPs/emails. Clean.

## Per-probe verdict

| Probe | Verdict | Why |
|-------|---------|-----|
| **`lrmagent.LRM`** | **KEEP** | Faithful to a strong paper, fully black-box reachable, and genuinely distinct (autonomy-over-techniques on a reasoning-model substrate — the paper's validated thesis, distinct from GOAT's prescribed menu). On-mission, client-actionable. This is the contribution of the batch — merge after the cleanups below. |

## What to do before merge

1. **De-dup plateau detection** into a shared `multiturn` helper used by Crescendo, `lrmagent`, and killchain.
2. **Add `lrmagent.LRM`** to the `--all-probes` skip list (`scan.go:643`).
3. **Bot fixes:** drop the stray `+` prefixes; `gofumpt`; apply the `AttackMaxAttempts` floor in `NewLRMWithGenerators` too.
4. **Make the GOAT contrast explicit** in the PR/docs; add a cost note and a short multi-turn escalation test.

- **DRY:** extract plateau detection to a shared helper (currently copied from `crescendo`).
- **Config:** add `lrmagent.LRM` to the unconfigured-multi-turn skip list so `--all-probes` doesn't run it with an empty goal.
- **Bot fixes:** stray `+` prefixes, gofumpt lint, `AttackMaxAttempts` floor parity in the test constructor.
- **Docs/tests:** explicit GOAT contrast, cost note, a multi-turn escalation test.

PR #142, opened 2026-06-01
build(release): emit SBOM + cosign signature (supply-chain hardening)

nsportsman:
## What

Backfills the **SBOM + cosign signing** blocks into this repo's `.goreleaser.yaml`, matching the proven julius config (PR praetorian-inc/julius#104).

```yaml
sboms:
  - artifacts: archive          # Syft SBOM per archive
signs:
  - cmd: cosign                 # cosign keyless OIDC signing of checksums.txt
    certificate: "${artifact}.pem"
    args: [sign-blob, --output-certificate=${certificate}, --output-signature=${signature}, ${artifact}, --yes]
    artifacts: checksum
    output: true
```

## Why

The central `go-release.yml` defaults `enable-sbom`/`enable-sign` true and installs Syft+cosign, but GoReleaser only emits these artifacts when the local config declares `sboms:`/`signs:`. Without them the supply-chain hardening was silently skipped — this repo's last release shipped no `.sbom`/`.sig`/`.pem`.

## Proven

julius `v1.2.1` (identical blocks) just shipped `checksums.txt.sig` + `checksums.txt.pem` + per-archive `*.sbom.json`. `goreleaser check` validated here too. Prereqs already satisfied by the workflow (Syft+cosign installed, `id-token: write` granted; cosign v2 keyless needs no COSIGN_EXPERIMENTAL).

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -28,3 +28,18 @@ changelog:
       - "^docs:"
       - "^test:"
       - "^chore:"
+
+sboms:
+  - artifacts: archive
+
+signs:
+  - cmd: cosign
+    certificate: "${artifact}.pem"
+    args:
+      - sign-blob
+      - "--output-certificate=${certificate}"
+      - "--output-signature=${signature}"
+      - "${artifact}"
+      - "--yes"
+    artifacts: checksum
+    output: true
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -32,6 +32,12 @@ changelog:
 sboms:
   - artifacts: archive
 
+# cosign keyless OIDC signing of the checksums file. NOTE: these are the cosign
+# v2 sign-blob flags. The central go-release.yml SHA-pins cosign via
+# sigstore/cosign-installer@v3.8.2 (=> cosign v2.4.3, verified shipping .sig+.pem).
+# cosign v3 replaces --output-certificate/--output-signature with a single
+# --bundle=${signature} (.sigstore.json); migrate this stanza in lockstep IF/when
+# the central workflow bumps cosign to v3.
 signs:
   - cmd: cosign
     certificate: "${artifact}.pem"


nsportsman:
Thanks @coderabbitai — verified the v3 facts independently, but this doesn't apply to our setup. Keeping the v2 stanza, with a clarifying comment added.

**The cosign v3 change is real** (confirmed via GoReleaser's [Upgrading to Cosign v3](https://goreleaser.com/blog/cosign-v3/) blog + cosign changelog): `--output-certificate`/`--output-signature` → required `--bundle=${signature}` (`.sigstore.json`).

**But we're pinned to cosign v2, not "newer/unpinned":** the central `go-release.yml` SHA-pins `sigstore/cosign-installer@3454372…` (`v3.8.2`), whose `action.yml` hardcodes `cosign-release: v2.4.3`. So every release deterministically runs **cosign v2.4.3**, where this exact stanza is correct.

**Empirically proven today:** julius `v1.2.1` (identical stanza) shipped `checksums.txt.sig` + `checksums.txt.pem` + per-archive `*.sbom.json` — run log shows `cosign-release: v2.4.3`. So "will fail" is not the case for the pinned version.

**Handling:** the real coupling is `signs:` stanza ⇄ the cosign major the workflow installs. Switching to `--bundle` now is unsafe (GoReleaser's blog doesn't confirm v2.4.x supports it, and it'd change verified-working output). Instead I've added an in-code comment documenting the v2 pin + the exact v3 migration; when the central workflow bumps cosign to v3, the installer pin + all capability `signs:` stanzas migrate to `--bundle` in one coordinated change. Not a per-PR blocker.

nsportsman:
Superseded by the cosign-v3 combined PR (https://github.com/praetorian-inc/augustus/pull/144) — that PR carries the SBOM + signing change in the modern v3 --bundle form plus the required go-release v2.5.0 pin.

PR #146, opened 2026-06-03
fix(ci): re-pin titus-scan to real tag v2.6.0 + adopt verify-pins gate

nsportsman:
## Why

Two pin-integrity fixes, in one PR per repo:

1. **Re-pin titus-scan to a real tag.** This repo pinned the Titus secret-scanner
   reusable with the comment `# v2.2.1` — but **that tag never existed** (public-workflows
   tags jump v2.2.0 → v2.3.0). A lying version comment defeats audits and Dependabot.
   Re-pinned to **`v2.6.0`** (`23254e12b026`), the current tag carrying the hardened,
   fail-closed, visibility-aware titus-scan reusable. Byte-identical scanner behavior —
   this is a label-honesty fix.

2. **Adopt the `verify-pins` gate.** Adds `.github/workflows/verify-pins.yml` calling the
   new reusable (also `v2.6.0`). On any future `.github/workflows/**` change it verifies
   every first-party public-workflows pin's `# vX.Y.Z` comment matches a real tag at that
   SHA — so a dishonest pin can never merge again.

All other first-party pins in this repo (go-ci, codex-code, claude-code, gemini-code,
claude-md-drift, external-contrib, go-sec) were audited and are already honest, so the new
gate passes on this PR.

### DIFF
--- a/.github/workflows/claude-code.yml
+++ b/.github/workflows/claude-code.yml
@@ -12,7 +12,7 @@ concurrency:
 
 jobs:
   claude-code-action:
-    uses: praetorian-inc/public-workflows/.github/workflows/claude-code.yml@debcee1fd3a78e908b1f89d8f2bef0644c454cbc  # v2.1.0  # v2.0.12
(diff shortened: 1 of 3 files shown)


nsportsman:
Superseded by a rebase onto current main (resolves claude-code.yml conflict from #145).

PR #204, opened 2026-06-24
Add WebSocket and hybrid HTTP+WebSocket generators

praetorian-farida:
## What

Adds two new generators (plus a shared helper package) for LLM front ends that don't speak plain request/response HTTP.

### `websocket.Websocket`
Generic `ws://` / `wss://` duplex generator. Sends a templated frame, reads the reply. Read modes: `single`, `until_close`, `until_marker`. Supports custom handshake headers, subprotocols, persistent connections (with transparent redial on a stale reused socket), rate limiting, and `wss` with optional `insecure_skip_verify`.

### `hybrid.Hybrid`
A **choreography engine** for targets that split the request and response across transports. The config is an ordered list of steps:

| step | purpose |
|---|---|
| `http` | templated request; `capture:` extracts vars; `answer:true`+`response_field` makes its body the reply |
| `http_poll` | re-request until a readiness condition (`until_field`/`until_value`, or answer-appears) — async job queues |
| `ws_connect` / `ws_send` / `ws_await` / `ws_stream` | open a socket, send frames, await a condition, assemble a streamed answer |

Values (a conversation id, job id, generated UUIDs) thread between steps via `capture` + `$NAME` substitution. This expresses **any** request/response transport pairing — HTTP-in/WS-out, WS-in/HTTP-out, pure HTTP, pure WS.

Motivating target: a GraphQL subscription chat where the prompt is an HTTP `updateConversation` mutation and the reply streams over a `graphql-transport-ws` subscription — auth in the `connection_init` payload, subscribe-before-prompt ordering (otherwise the backend 500s), and `CompleteSignal`-terminated stream assembly.

Also supports: alternative-schema fallback (multiple `forms`, `response_field` lists), persistent multi-turn sessions (`once:true` setup steps), an optional `reuse_connection` toggle, and an optional CONNECT `proxy` (Burp).

### `wsutil`
Shared dial/handshake, frame, and JSON-field helpers used by both generators (no duplication).

## Notes
- Reuses the existing `golang.org/x/net/websocket` dependency — **no new modules**, `go.mod`/`go.sum` untouched.
- Registered in `pkg/register/generators`.

## Testing
- Per-package tests incl. a full in-process HTTP+`graphql-transport-ws` harness replicating the subscription-chat flow, transport-matrix cases (pure HTTP, pure WS, hybrid), fallback, polling (until-field / answer-appears / exhaustion), and config validation.
- `go test -race`, `go vet`, and `golangci-lint` all clean; whole repo builds.


maurodelossantos-praetorian (approved):
LGTM! Two VERY minor things my local Claude flagged and some optional follow-ups:

1. Warn when TLS verification is disabled. `wsutil.BuildHandshakeConfig` and `hybrid.NewHybrid` set `InsecureSkipVerify` with no log line; the REST generator warns at `rest.go:62`.
2. Cover the CONNECT-proxy path. `DialViaProxy`/`EnvProxyFor`/`readCONNECTStatus` are at 0% test coverage but the Burp/HTTPS_PROXY flow is a primary use case. A fake-proxy listener test (plus a wss TLS-over-tunnel case and a 407/502 rejection) would lock it in.

Optional follow-ups, not blocking: `idleDeadline` returns an always-true bool (hybrid.go:391); the header comment says `websocket.Hybrid` but the type is `hybrid.Hybrid`; and a hybrid hook-var test would mirror the websocket one.

praetorian-farida:
Thanks @maurodelossantos-praetorian — addressed in fd70215:

1. **Warn on disabled TLS verification.** `wsutil.BuildHandshakeConfig` (wss leg) and `hybrid.NewHybrid` (HTTP leg) now `slog.Warn` when `insecure_skip_verify=true`, mirroring the REST generator.
2. **CONNECT-proxy coverage.** Added a fake-proxy listener test covering the ws tunnel, the **wss TLS-over-tunnel** case, and **407/502 rejection**, plus `EnvProxyFor` env mapping (HTTPS_PROXY/HTTP_PROXY/NO_PROXY + explicit override). Coverage on the previously-0% path: `DialViaProxy` 74%, `EnvProxyFor` 90%, `readCONNECTStatus` 88%, `connectSucceeded` 100%.

Optional follow-ups also done:
- `idleDeadline` now returns a bare `time.Time` (the bool was always true).
- Fixed the `hybrid.go` header comment (`hybrid.Hybrid`, not `websocket.Hybrid`).
- Added a hybrid hook-var substitution test mirroring the websocket one.

Race/vet/lint clean across all three packages.

PR #210, opened 2026-07-02
feat: MCP generator, direct tool-surface probes, and first-class reconnaissance

praetorian-farida:
Adds first-class support for testing **MCP (Model Context Protocol) servers** in augustus: a generator that speaks MCP, security probes that call the tool surface directly, and a new **reconnaissance** module family that gathers target facts as observations (not verdicts) and can feed them to probes.

### 1. MCP generator (`internal/generators/mcp`)
- Transports: `stdio`, streamable `http`, legacy `http+sse`, and `auto` detection.
- Implements `types.ToolInvoker` (`ListTools`/`CallTool`) and `types.MCPReconnaissance` (full attack-surface inventory) via the official MCP Go SDK.
- Explicit `proxy` config key (Burp-friendly); session context kept alive for the whole session (SSE lifetime fix + regression tests).

### 2. Direct tool-surface probes (`internal/probes/toolsec`)
Unlike the model-facing probes (which test whether an *LLM* can be manipulated), these call real tools directly and test the tool backend:
- `toolsec.Injection` — computed-arithmetic canary payloads detect code/command/template injection sinks with no echo false-positives.
- `toolsec.SSRF` — out-of-band canary URLs detect blind + reflected SSRF.

### 3. First-class reconnaissance (`pkg/recon`, `pkg/output`, `internal/recon/mcp`)
Recon **measures** (gathers descriptive data); it does not render a verdict. Forcing it through the scored-probe pipeline was the source of a lot of friction, so it's now its own thing:
- `pkg/output.Observation` — the one descriptive output type (`Type`/`Target`/`Data`/`Source`). The verdict stays the score; it is not re-represented.
- `pkg/recon` — a `Recon` interface that returns `[]Observation` (so it *structurally* cannot carry a verdict), a registry (6th module family), a concurrency-safe `Store`, and a best-effort runner.
- `recon.MCP` — enumerates the MCP attack surface into one `mcp.inventory` observation.
- `cmd/augustus` — a `--recon` flag and a recon phase that runs independently of the detector harness (recon-only scans skip it entirely).

### 4. Recon feeds probes (Metasploit model — scan once, reuse everywhere)
- `recon.ProbeContext` + `ContextAwareProbe`: an **opt-in** interface, so `Prober` is unchanged and probes that don't opt in structurally cannot see recon.
- `toolsec.Injection`/`SSRF` consume the shared inventory and only fall back to a live `ListTools` when none is present.

> Note: an earlier commit added recon as a scored probe (`toolsec.Recon`); this PR **supersedes and removes it** — recon is no longer a verdict. Tool-poisoning *verdict* work is deferred to a separate test probe (LAB-4460).

## Verification
- `go build ./...`, `go vet`, `gofmt`/`gofumpt`, `golangci-lint` all clean; full suite (168 packages) passes.
- End-to-end against DVMCP: `recon.MCP` populates the store, the probe tests the recon-discovered tools, catches a real command-injection sink on challenge8 (`857367*264` → `Result: 226344888`), and correctly reports no false positive on challenge2.


praetorian-farida:
**Hook + MCP interplay (the important ones):**
- **HookedGenerator dropped MCP interfaces** — `NewHookedGenerator` now returns a *capability-preserving* wrapper that exposes `ToolInvoker`/`MCPReconnaissance` **only when the inner generator implements them**. A hooked MCP scan now runs recon + tool-surface probes; a hooked plain generator is *not* falsely advertised as MCP-capable (which would break the type-assert-and-skip pattern). The prepare-hook + var-injection prologue is extracted to a shared `hookContext`, so hook-provided credentials reach every call path (Generate, ListTools, CallTool, MCPInventory).
- **Header hook-vars not substituted per request** — `headerTransport` now substitutes `$KEY` + `$VARNAME` from the **per-request context** instead of baking headers once at client build, matching the REST generator (longest-key-first). `Authorization: Bearer $TOKEN` from a prepare hook now resolves.

**Correctness:**
- **Cleanup hooks on recon-only scans** — extracted `runCleanupHook`, now invoked on both completion paths so a `--setup`-provisioned resource is torn down even when no probes run.
- **Recon banner breaks JSONL stdout** — banner moved to stderr; stdout stays a pure JSONL stream.

**gosec:**
- **G204 (stdio subprocess)** and **G402 (InsecureSkipVerify)** — both are operator opt-ins (stdio launches an operator-configured MCP server; `insecure_skip_verify` is for self-signed scan targets). Annotated with `#nosec` + justification, matching the repo convention (the `go-sec.yml` workflow honors `#nosec`; golangci doesn't run gosec here).

**One I believe is a false positive:**
- **"Decode MCP input schemas before exposing tool parameters"** — on the *client* path this doesn't apply: the SDK's `Tool.InputSchema` is `any` with `json:"inputSchema"`, so a received `tools/list` decodes it into `map[string]any` — exactly what `toolParams` needs. Verified with a real-SDK-over-httptest regression test (`TestToolInvoker_ListTools_ParametersDecoded`) that asserts `parameters` is a decoded map with `properties`. (The typed `*jsonschema.Schema` only exists when *constructing* a Tool server-side, which the generator never does.) Kept the test as a guard.

All changes are TDD'd; full suite (179 packages) + `go vet` + `golangci-lint` pass.

praetorian-farida:
**Update on gosec G204 (stdio command exec):** superseding my earlier comment — rather than keep the stdio transport with a `#nosec` suppression, we **removed the stdio transport entirely** (`8505df1`).

Reasoning: stdio only reaches a *local* subprocess, so it has no role in scanning *remote* MCP targets (which use http/sse). And when augustus runs as a hosted capability, `exec.Command(cfg.Command, …)` is an arbitrary-code-execution surface if any config field is ever driven by untrusted input — naming the program to run *is* the exploit, so no allowlist/validation on the command string can make it safe. Removing the transport eliminates the exec, so the G204 finding is gone at the root (no suppression to defend). Supported transports are now http/sse/auto — all network, none exec.

maurodelossantos-praetorian (changes requested):
# Review: MCP generator, tool-surface probes & recon

First off: **great work, and I agree with the design.** This is well implemented and the right foundational basis for the MCP testing capability. Before the line-level findings, I want to frame how I see this upgrade, because it changes how we read "what's done vs. what's next."

## 1. Design & architecture: LGTM on all of it

The core call is sound. Modeling the MCP server through the `Generator` interface with `ToolInvoker` / `MCPReconnaissance` as optional interfaces (gated exactly like `VisionCapable`/`DocumentCapable`) keeps `Prober` and the scan pipeline untouched, and the first-class `recon` module family (`Observation` + `Store` + opt-in `ContextAwareProbe`, the "scan once, reuse everywhere" model) is a clean separation of *measuring* from *verdict*. The seams are well chosen and forward-compatible. I don't think any of this needs rework. Removing `stdio` was the right security call too.

## 2. How I'm framing this upgrade (so we're aligned on scope)

**Two distinct modes, and this PR is the foundation for one of them.**

- **Mode A (direct testing).** We reach the MCP server directly (assuming internal/authorized access) and drive its tools ourselves, no LLM in the loop: classic DAST over the MCP wire. **This PR is the foundation for Mode A**, and it's a solid one. Mode A is *not finished* by this PR, though: it deliberately ships the spine (generator + recon + two flagship probes: injection & SSRF). Still ahead for full Mode A coverage: authorization/BOLA, path traversal (the LAB-4462 metrics put this at ~82% prevalence), blind/out-of-band injection, the resources/prompts surface (enumerated but not yet probed), secret-leakage, and protocol-level fuzzing. That's expected follow-up, not a gap in this PR.
- **Mode B (agent-mediated testing).** The realistic "attacker → LLM/agent → MCP tools" scenario (tool-description poisoning, indirect prompt injection, rug pulls, cross-server shadowing). **This is a different thing and out of scope here.** The good news is this PR positions it well: Mode B should be built as a *new composition over `ToolInvoker`* (generator = the LLM, MCP = a tool provider attached to the conversation), reusing the recon `Store` and inventory, not as an extension of the `mcp.MCP` generator. Worth keeping that line as we build B.

**Julius enumeration vs. Augustus enumeration: different layers, not overlap.**

- **Julius** answers *"is this an MCP server, and how is it locked?"* That is identification only (open vs. OAuth/RFC-9728), and it does **not** enumerate the tool catalog.
- **Augustus** (`recon.MCP`) answers *"what does it expose?"* It connects as an MCP client and enumerates tools/resources/prompts/capabilities, then tests them.

So Julius hands over `{endpoint, transport, auth-state}` and Augustus does the deep work. No duplication. This PR already anticipates that handoff (the `Store.Observe` ingestion API + per-observation `Source`), so wiring Julius's output in later is additive.

## 3. Findings from my review (verified locally)

These come from a multi-agent + multi-model pass (three specialist review agents plus the CI reviewers), each finding verified against the code. Grouped by whether it gates merge.

### Blockers (fix before merge)
- **F1: retry re-invokes a tool on request timeout.** `withSession` gates its reconnect-and-retry on the parent `ctx`, but the per-call `RequestTimeout` lives on an inner `callCtx`, so a slow `tools/call` reconnects and **calls the tool a second time**, a duplicated side effect against real infrastructure. Guard with `!errors.Is(err, context.DeadlineExceeded)` (and `context.Canceled`). *(mcp.go:179)*
- **F2: add the regression test for F1.** No test currently fails if the double-invoke guard is removed (the fall-through path is uncovered). Pin it.
- **F3: recon-only scan is a false-green.** Recon errors are logged-and-swallowed and a recon-only run returns success, so `--recon` against an unreachable server exits 0 with zero observations. Exit non-zero when recon was the only activity and it errored/produced nothing. *(scan.go)*
- **F4: credentials re-injected across redirects.** `headerTransport.RoundTrip` sets `Authorization`/`api_key` on every hop with no host check; a malicious target can 302 to its own host and harvest the token (Go's cross-host stripping doesn't cover headers added by a custom `RoundTripper`). Set `CheckRedirect: http.ErrUseLastResponse` or host-gate the injection. *(mcp.go:647)*
- **F5: direct tool execution has no safety gate.** The probes call real tools with adversarial args, and the MCP destructive-hint annotations are dropped in `toolsToMaps`/`mcpToolsFrom`, so there's no way to skip destructive tools even if we wanted to. Fine for DVMCP; risky against customer infra. Proposal: capture the annotations and default to read-only/allow-listed tools behind an explicit opt-in flag (+ tracking issue) rather than hard-gating this PR.
- **Rebase:** branch currently conflicts with `main`.

### Should-fix (tests + one design item)
- **F6: `HookedGenerator` capability preservation is incomplete.** The per-combination wrapper covers only `ToolInvoker`+`MCPReconnaissance`; it drops `VisionCapable`/`DocumentCapable`/`RawResponseProvider`, so a multimodal probe run *with a hook* mis-reports the target as unsupported. Pre-existing, but this PR reworks exactly this file, so worth fixing here (or an immediate follow-up + issue) with an interface-count-agnostic approach.
- **F7: real `MCPInventory` enumeration is 0% tested** (`recon.go`), and it anchors the recon→toolsec chain, so add an httptest/SDK-server test.
- **F8: `insecure_skip_verify` + proxy wiring untested** in the MCP path.
- **F9: session reconnect happy/fail paths only incidentally covered.**

### Non-blocking comments
- **F10:** recon inventory records `Transport:"auto"` instead of the resolved `http`/`sse`; use `DetectedTransport()`. *(recon.go:27)*
- **F11:** `toolsec.Injection` silently returns `nil` on a non-`ToolInvoker` target (reads as clean); align with the fail-loud convention.
- **F12 (please verify):** under concurrent probes sharing one generator, a failing call's `dropSession()` may tear down the session another probe is mid-call on. Couldn't confirm this locally, so worth a check.
- Nits: unbounded tools×params×payloads amplification (optional cap), `substituteHeader` sequential-replace, and a few small test gaps (SSRF knobs, hooks branch selection, MCPInventory error path).

maurodelossantos-praetorian (commented):
## Re-review of the latest changes (through `374f717`)

I re-verified the fixes against the code and ran the touched packages' tests locally, all green. Every blocker and should-fix from the previous round is addressed and covered by a test.

From my side the code is merge-ready. One non-blocking recommendation remains.

## Recommendation (non-blocking): update CLAUDE.md to clear the drift advisory

The `drift-check` advisory is accurate: this PR adds a new module family (`recon`) and two new optional generator interfaces that the architecture doc doesn't mention yet. CLAUDE.md isn't part of this PR's diff, so I can't attach a click-to-commit suggestion.

Everything else looks good. Nice foundation.

maurodelossantos-praetorian (approved):
LGTM! I left one minor comment, completely optional, about a potential documentation update.

PR #228, opened 2026-07-07
feat(multimodal): add three paper-faithful image-based attack probes (LAB-2417)

maurodelossantos-praetorian:
## Summary

Adds paper-faithful **image-based attack probes** for [LAB-2417](https://linear.app/praetorianlabs/issue/LAB-2417). After a critical re-review of the original six against their source papers, this PR ships the **three that faithfully reproduce their attack on the image path** and moves the rest to dedicated tickets (they need a harness/modality Augustus's single-turn image path doesn't have).

### Probes shipped (3)

| Probe | Technique | Reference |
|---|---|---|
| `multimodal.FigStepNumberedList` | Instruction rendered as a numbered list — the faithful FigStep mechanism | arXiv:2311.05608 (AAAI'25) |
| `multimodal.VSHNarrativeScene` | Narrative-hypnosis **text prompt** + operative instruction in the **image** (the paper's modality-asymmetric mechanism) | Pattern Recognition 2025/26 |
| `multimodal.VisualSemanticExploits` | Emoji/icon **rebus** (🖨️👋🌍 → "print hello world"), no literal text | NVIDIA "Semantic Prompt Injections" |

All are `Covert=false`, scored by `multimodal.Canary`, and are **mechanism-faithful with a benign canary** (they reproduce the attack mechanism, not the papers' harmful-content ASR).

### Cut here, tracked as follow-ups (under LAB-2364)
- **HomoglyphOCRConfusion** — homoglyph code points vanish on rasterization → identical to `InstructionImage` against a pixel-only VLM; the technique is already covered in the text channel by `badchars.Homoglyphs`. Removed.
- **MaliciousFontInjection** — the render-vs-codepoint divergence can't survive a flat PNG; faithful only via the document harness → **LAB-4881**.
- **VisualContextualAttack (VisCo)** — is photorealistic multi-image + fabricated multi-turn dialogue, not rendered text tiles → **LAB-4882**.

## Detector changes (`multimodal.Canary`)
- `AllCanaries`: FigStep `AMBER VORTEX 7042`, VSH `ZIRCON SAFARI 1764`, Semantic `hello world` (an *imageable* phrase — a rebus can't encode an opaque token; it leans on the `matchesCanary` distance gate rather than the p-value decoy test).
[...]

PR #230, opened 2026-07-08
fix(anthropic): make temperature opt-in so newest models work (LAB-4691)

maurodelossantos-praetorian:
Linear: LAB-4691

## Summary

The Anthropic generator hardcoded a default `temperature: 0.7` and sent it on **every** request. The newest Claude models — `claude-sonnet-5`, `claude-opus-4-x` — reject the field outright:

```
anthropic: bad request (invalid_request_error): `temperature` is deprecated for this model.
```

So the generator could not target the current Claude generation at all without a manual `-c '{"temperature":0}'` workaround (which only worked because `omitempty` dropped the zero).

## Fix

Make `temperature` **opt-in** by switching `Config.Temperature` and the request wire field to `*float64`, mirroring the existing `ollama` generator:

- **nil (unset)** → field omitted from the request → works against every model, including those that reject `temperature`.
- **explicitly configured** (via `-c '{"temperature":…}'`, config file, or `WithTemperature`) → sent as-is, **including an explicit `0`** — which the previous `float64` + `omitempty` encoding could not represent (it conflated "unset" with "0").

## Behavior change

A scan that does **not** set temperature no longer injects `0.7`; the provider's own default applies. This is the intended correction — the scanner should not inject a sampling parameter the operator never asked for, and injecting one made the newest models unusable.

## Tests

- `TestAnthropicConfigDefaults` — temperature is `nil` by default.
- `TestAnthropicGenerator_DefaultTemperature` — rewritten to assert the field is **absent** from the request when unset.
- `TestAnthropicGenerator_ExplicitZeroTemperature` — **new**: explicit `temperature: 0` is honored and sent.
- Config-from-map / options tests updated for the pointer type.

## Verification

- `go build ./...`, `go vet`, `gofmt`, `go test -race ./internal/generators/anthropic/` and `./pkg/config/` all green.
- **Live:** `augustus scan anthropic.Anthropic -m claude-sonnet-5 --probe dan.Dan_11_0 --detector dan.DAN` (no temperature) now returns a real response (`error: none`), where it previously failed with the deprecation error.

## Context

Surfaced while validating the image-probe work in [LAB-2417](https://linear.app/praetorianlabs/issue/LAB-2417) (PR #228) against live vision models — the `temperature` rejection blocked testing any probe against the newest Claude models. This is an independent generator fix, kept in its own PR.

### DIFF
--- a/internal/generators/anthropic/anthropic.go
+++ b/internal/generators/anthropic/anthropic.go
@@ -32,11 +32,10 @@ func init() {
 
 // Default configuration values matching litellm patterns.
 const (
-	defaultMaxTokens   = 150
-	defaultTemperature = 0.7
-	defaultAPIVersion  = "2023-06-01"
-	defaultBaseURL     = "https://api.anthropic.com/v1"
-	defaultTimeout     = 90 * time.Second
+	defaultMaxTokens  = 150
+	defaultAPIVersion = "2023-06-01"
+	defaultBaseURL    = "https://api.anthropic.com/v1"
+	defaultTimeout    = 90 * time.Second
 )
 
 // Anthropic is a generator that wraps the Anthropic Messages API.
@@ -48,8 +47,9 @@ type Anthropic struct {
 	apiVersion string
 	model      string
 
-	// Configuration parameters
-	temperature   float64
+	// Configuration parameters. temperature is nil when unset so it is omitted
+	// from the request (see Config.Temperature).
+	temperature   *float64
 	maxTokens     int
 	topP          float64
 	topK          int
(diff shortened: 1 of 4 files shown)


maurodelossantos-praetorian:
Reviewed the AI feedback (Codex, Claude, Gemini) and addressed it in `bb1c1b1`.

**Fixed:**
- **Type-safe temperature parsing** (Gemini): `ConfigFromMap` checked key *presence* then `GetFloat64(…, 0)`, so a present-but-malformed value (e.g. the string `"0.5"`) was silently coerced to an explicit `0` and sent — forcing unexpectedly deterministic generation. Now it parses via a type switch: the `*float64` is set only for a real number (`float64`/`int`); a missing key **or** a malformed value leaves it `nil` (unset → omitted), matching the ollama precedent. Explicit `0.0` is still honored. Added `TestAnthropicConfigFromMap_TemperatureTypeSafety` (string/bool/absent → nil; int/explicit-0 → set).

**Not fixed (intentionally, out of scope):**
- **Pointer-ify `TopP`/`TopK`** (Gemini): valid future consistency point, but it's beyond this temperature-focused change and an explicit `0` for those is rarely needed. Left for a follow-up.

**No action needed:** Codex and Claude both returned LGTM with no critical issues (Claude confirmed the `*float64` + `omitempty` encoding is correct — nil omits, explicit `0` serializes).

`go build`, `go vet`, `gofmt`, and `go test -race ./internal/generators/anthropic/` all pass. This PR is good for review.

praetorian-farida (approved):

PR #233, opened 2026-07-10
feat: MCP BOLA probe + first-class recon composition (recon.MCPIdentifiers, toolsec.BOLA)

praetorian-farida:
MCP **Broken Object-Level Authorization (BOLA)** testing for Augustus, plus the first-class recon composition it needs. Rebased onto `main` (now that #210 has merged).

## Commits
1. **recon composition foundation** — `recon.ContextAwareRecon` (a recon module can consume earlier observations; the runner injects the shared store), a `recon.settings` config section + `ResolveReconConfig` (with env-var interpolation, incl. list recursion), and `internal/recon/llm.Base`, embeddable navigator-LLM plumbing (lazy generator, reuses the shared judge).
2. **`recon.MCPIdentifiers`** — establishes BOLA *ownership ground truth*: an LLM navigator classifies getter/enumerator tools (deterministic operator-editable keyword fallback), enumerates each identity's own object ids, round-trip-validates them, and emits `mcp.identifiers` observations of `{Tool, Param, ID, Args, Source}`. Ownership is the enumeration **set-difference** across identities — not response parsing.
3. **`toolsec.BOLA` probe + detector** — the probe is a pure payload sender (ATTACK `get(X)` + a nonexistent-id NEGATIVE control + an own-object POSITIVE control). The detector is a **server-agnostic prune→judge chain**: a deterministic prune (target response == nonexistent-id response, both ids masked → denied) then an opt-in, control-calibrated LLM judge (leak / denied / uncertain). **No fingerprint hashing, no ownership-marker/field-name guessing, no id-format assumptions.** No judge configured → `0.5` inconclusive + warn (never a silent false negative).
4. **shared `toolpolicy`** — extracts #210's destructive-tool safety gate into `internal/toolpolicy` and routes recon, BOLA, injection, and ssrf through it. recon filters by MCP `read_only`/`destructive` **annotations**, not a name regex.

## How it runs
`recon.MCP` (inventory) → `recon.MCPIdentifiers` (per-identity harvest + validation) → `toolsec.BOLA` (attack + controls) → detector (prune → judge). Identity travels as a request header (a second identity is intrinsic to BOLA and is configured).

## Why the detection is generic (not overfit)
Ownership comes from recon's set-difference (the "business context" generic scanners lack, per Stingr.ai/Unit42). Served-vs-denied is decided by comparing the attacker's response to *its own* nonexistent-id baseline, and the judge is calibrated by the P/N/T controls — so it works on any response format. Detector unit tests exercise multiple server dialects (JSON not-found, plain text, empty-`{}`, "403 forbidden", numeric ids).

## Validation
End-to-end against a purpose-built multi-tenant DVMCP BOLA challenge (streamable HTTP, header tenant auth, `list_orders` scoped / `get_order` vulnerable), with a real **Haiku** navigator *and* judge: recon harvests both tenants' ids, and `toolsec.BOLA` scores **1.0** on tenant-a reading tenant-b's order via the judge. Full gauntlet: `build` · `vet` · `golangci-lint` (0) · `go test ./...` — all green.

## Review history
Evolved through Codex + a human review: the initial fingerprint/ownership-marker detection was replaced with the server-agnostic prune→judge chain above; the destructive-tool guard moved to the shared `toolpolicy`; victim `${VAR}` interpolation, replay-args, `(tool,id)` id-scoping, judge-cred reuse, and identity-label validation are all addressed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -47,7 +47,7 @@ Generators may also implement these **optional** interfaces (in `pkg/types/gener
 - **UsageReporter**: `AccumulatedTokens() int64` — reports the cumulative tokens consumed across all `Generate` calls. The scanner type-asserts each generator for this interface and records the per-run delta into `Metrics.TokensConsumed` (surfaced as `augustus_tokens_consumed`). Implement it for free by embedding `types.UsageCounter` (a concurrency-safe `atomic.Int64`) and calling `g.AddTokens(...)` wherever the provider returns usage. Generators whose provider doesn't report usage still embed `UsageCounter` and simply never `AddTokens`, contributing 0 (honest partial coverage, never an estimate).
 - **VisionCapable**: `SupportsVision() bool` — declares that the generator's wire layer can transmit image attachments (`Message.Images`). Multimodal image probes gate on this to skip generators that can't carry images rather than silently sending a text-only request and mis-reporting the target as safe. Report **structural** capability (the generator emits image content blocks), not per-model support — e.g. an OpenAI-compatible generator returns `true` on its chat path even though a given model may ignore images; for generators with both image-capable and text-only paths (OpenAI/Azure completion models, Bedrock Titan/Llama), return the path-accurate value (e.g. `g.isChat`, or the model family).
 - **DocumentCapable**: `SupportsDocuments() bool` — the document-modality parallel of `VisionCapable`: declares that the generator's wire layer can transmit document attachments (`Message.Documents`, e.g. PDFs). Document probes (`internal/probes/pdf/*`) gate on this so a generator that can't carry documents fails the probe rather than silently sending a text-only request and mis-reporting the target as safe. Report **structural** capability — Anthropic returns `true` unconditionally; Bedrock returns `true` only for the Claude builder (Nova/Titan/Llama return `false`, as only the Claude path emits document blocks).
-- **ToolInvoker** (`pkg/types/tool_invoker.go`): `ListTools()` / `CallTool()` — declares that the target exposes a directly-invokable tool surface (e.g. an MCP server) rather than only chat completion. This is distinct from the model-facing tool wire layer (`Conversation.Tools`/`Message.ToolCalls`), which presents probe-defined tools *to* an LLM and executes nothing: `ToolInvoker` invokes REAL tools on the backend. It is the basis for the `internal/probes/toolsec/*` probes (authorization, injection-into-a-sink, SSRF against tool backends).
+- **ToolInvoker** (`pkg/types/tool_invoker.go`): `ListTools()` / `CallTool()` — declares that the target exposes a directly-invokable tool surface (e.g. an MCP server) rather than only chat completion. This is distinct from the model-facing tool wire layer (`Conversation.Tools`/`Message.ToolCalls`), which presents probe-defined tools *to* an LLM and executes nothing: `ToolInvoker` invokes REAL tools on the backend. It is the basis for the `internal/probes/toolsec/*` probes (broken object-level authorization via `toolsec.BOLA`, injection-into-a-sink via `toolsec.Injection`, and `toolsec.SSRF` against tool backends).
 - **MCPReconnaissance** (`pkg/types/mcp_recon.go`): `MCPInventory()` — declares that the target's full MCP attack surface (declared capabilities, negotiated protocol version, server instructions, and the tool/resource/prompt catalog) can be enumerated from the connected session. Implemented by the `mcp.MCP` generator and consumed by the `recon.MCP` reconnaissance module. Assembles raw descriptive data only — it renders no verdict.
 
 ### Reconnaissance (pkg/recon/)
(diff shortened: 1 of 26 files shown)


maurodelossantos-praetorian (changes requested):
# Review: MCP BOLA probe + first-class recon composition

Strong, well-architected foundation. Almost everything below is about the plumbing *around* the core logic, where a scanner's worst outcome, reporting a vulnerable server as safe, can slip through.

A note on attribution: several of these were also flagged by the Codex auto-review (marked "Codex P2"); our review independently reached the same conclusions and adds the rest.

## Blockers (fix before merge)

The common thread across B1 to B5: a genuinely vulnerable multi-tenant server can be scanned and reported clean.

- **B1: victim `${VAR}` interpolation is broken (Codex P2, and confirmed by our review).** `interpolateMapEnvVars` (`pkg/config/loader.go`) recurses into maps and strings but has no `[]any` case, so `${...}` placeholders inside `victims[].generator_config` (API keys, tenant tokens) stay literal. The second identity fails to authenticate, harvests zero object ids, and BOLA silently emits no attempts, a false "safe". A code comment claims victim configs are interpolated, so this also contradicts its own docs. Add an `[]any` case plus a slice-nested interpolation test.
- **B2: BOLA replay drops required non-id arguments (Codex P2).** `recon.MCPIdentifiers` validates a getter with `benignArgs` (filling every other required param), but the replay in `bola.go:113` sends only the id param. A getter that also requires `type`/`scope`/`format` returns `IsError`, and the detector scores it safe even though recon already confirmed the object is readable with those arguments. Replay with the same argument set recon validated with.
- **B3: own-id suppression is keyed on the raw id, not (tool, id) (Codex P2).** `bola.go:76-92` builds the attacker-owned set from bare id strings, so if the attacker owns any object with id `1`, every victim object with id `1` is skipped, even across different object classes (`get_order` vs `get_ticket`). `MCPObjectRef` already records the tool and param; scope the suppression to the same getter.
- **B4: default detection collapses to exact fingerprint match.** The "id plus distinguisher" path is inert unless the operator configures `owner_markers`, and `normalize()` (`identifiers.go`) deliberately keeps volatile fields, so any getter response carrying a timestamp or request id produces a different hash on replay and a real leak scores 0.0. Require a secondary signal, use tolerant fingerprinting, or at minimum warn when exact-match is the only active signal.
- **B5: the `IsError` short-circuit trusts a target-controlled flag.** The detector (`detectors/toolsec/bola.go:54`) returns safe the moment `IsError` is set, regardless of body, so a server that returns the victim object inside an error-shaped envelope is scored safe. Fingerprint-check the body before concluding safe.

## Should-fix

- **S1: no safety gate on tool invocation.** Neither the recon harvest nor the BOLA replay applies #210's `toolPolicy`, and the heuristic classifier treats any tool with no required id-like param as an enumerator and calls it (`identifiers.go:277`), with `benignValue(bool)` returning `true`, so recon can invoke a `reset_*`/`delete_*`-style tool by default. On rebase, route both paths through `toolPolicy` and honor read-only/destructive annotations. (The same gap exists in #234's PathTraversal, so it is worth handling once across the toolsec family.)
- **S2: the navigator never receives the shared judge credentials (Codex P2).** `ResolveReconConfig` injects the shared judge model/key under `judge_config`, but `NewBase` only reads `navigator_config` and top-level model keys, so in the documented shared judge/navigator setup the navigator silently falls back to heuristics or provider defaults unless the operator duplicates the config.
- **S3: fingerprint normalization is duplicated** in `identifiers.go` and the detector, kept in sync only by a comment. A fix to B4 on one side would silently break detection on the other. Extract one shared helper.
- **S4: cross-tenant data at rest.** The raw victim object body is written to the BOLA attempt output and victim ids into observations, a secondary disclosure of the exact data the finding is about. Store a fingerprint plus a minimal redacted snippet instead.
- **S5: the two identity labels must agree.** `attacker_identity_label` (probe config, default `primary`) and `identity_label` (recon config, default `primary`) live in different config sections with no cross-check; if they drift, the attacker replays its own objects against its own session and scores a false positive. Validate that the attacker label matches a discovered identity.
- **S6: no end-to-end probe to real-detector test.** The probe test re-implements scoring in a stub that diverges from the real detector exactly on the `IsError` branch, so a metadata key/type drift between probe and detector would pass both suites yet score 0.0 in production. Add one test that runs a real attempt through the real detector, plus coverage for numeric ids, harvest failure, and the owner-marker path.

## Non-blocking comments

- The navigator ships the target's full tool catalog (names, descriptions, schemas) to OpenAI by default (`use_navigator=true`, inheriting the judge key). Worth surfacing as an explicit consent / data-handling decision rather than a silent default.
- recon to recon ordering is convention-only; it degrades safely to a live `ListTools` if modules run out of order, but a declared dependency or an ordering warning would harden it.
- `benignValue` and the tool-schema parsing are duplicated between `recon/mcp` and `probes/toolsec`; and BOLA returns `nil` silently on a non-`ToolInvoker` target, the fail-loud pattern the other toolsec probes adopted in #210.

## Ticket and housekeeping

- **Tracking ticket.** Filed LAB-4879 under the MCP epic (LAB-4458) for this work and linked it to this PR.
- **Merge conflict.** The branch currently conflicts with `mcp-generator` in `cmd/augustus/scan.go`, `pkg/config/config.go` (+ test), and `pkg/types/mcp_recon.go` (+ test), which are the files #210's review-fix commit also touched. Rebase onto the updated base to clear it. (It is blocked on #210 until that lands anyway.)

maurodelossantos-praetorian:
**Follow-up (optional, your call @praetorian-farida ):** once the feedback above is addressed (and #210 is merged), it may be worth running the standard CI review agents (Codex, Claude, Gemini) over the updated code, so this PR goes through the same complete CI/CD review process we apply to the others. Any of them can be invoked on demand by posting a comment that @-mentions the reviewer by name on one of the lines of code. Entirely deferred to you.

maurodelossantos-praetorian (approved):
LGTM!

PR #234, opened 2026-07-10
feat(toolsec): path-traversal, DNS-rebinding, SSE session-hijack probes (LAB-4462)

praetorian-farida:
## Summary

Extends the `toolsec` family from PR #210 with three of the four remaining LAB-4462 categories. Base is `main` (rebased 2026-07-14 after #210 merged).

**New probes / detectors**:

- `toolsec.PathTraversal` — injects OS-file traversal payloads into path-like tool parameters and detects reads via file-content signatures (Linux `root:x:0:0:`, macOS `# User Database` + `nobody:*:`, Windows `[extensions]`). Extracts allowed prefixes from the tool description ("Only files in /tmp/safe/") and generates prefix-append payloads that defeat `startsWith` gates.
- `toolsec.DNSRebinding` — sends randomised RFC 2606 Origin/Host values that a spec-compliant validator must refuse, plus a CORS preflight looking for credentialed reflection. Payloads randomised per run so no substring-blocklist WAF matches. Classifies findings by concrete validator weakness (external-origin, null-origin, extension-origin, localhost-lookalike, case-variant, unexpected-host, cors-reflect-creds).
- `toolsec.SSESessionHijack` — samples N session IDs and inspects for length / entropy / prefix / collision weaknesses, then tests whether the server accepts a sampled session id from a fresh TCP connection or after the stream closed. Includes a control test that ensures the server rejects fabricated ids, and (this PR's `cde70ac` fix) suppresses the connection-lifetime replay findings when an outbound proxy is configured — those tests are proxy-artefact prone.

**Interface change** (`pkg/types.MCPEndpoint`): gains `HTTPClient() *http.Client` and `ProxyURL() *url.URL`. The transport-layer probes now borrow the generator's HTTP client instead of building their own — proxy, TLS, and injected headers flow through automatically. `ProxyURL()` is the escape hatch for probes that make claims sensitive to persistent-connection semantics.

## LAB-4462 coverage status after this PR

| # | Probe | Status |
| :-- | :-- | :-- |
| 1 | SSRF | ✅ #210 (`toolsec.SSRF`) |
| 2 | Command injection | ✅ #210 (`toolsec.Injection`) |
| 3 | Path traversal | ✅ this PR |
| 4 | DNS rebinding | ✅ this PR |
| 5 | SSE session hijacking | ✅ this PR |
| 6 | OAuth 2.1 proxy flaws | ⬜ separate PR (needs DCR + `.well-known/oauth-authorization-server` walk) |

## Verification

Live against DVMCP (all 10 challenges up on :9001–:9010):

| Probe | Target | Result |
| :-- | :-- | :-- |
| `toolsec.PathTraversal` | ch2 `read_file` | ✅ 1 finding — `/tmp/safe/../../../../../../etc/passwd` returned real /etc/passwd (`root:*:0:0:System Administrator:…`) |
| `toolsec.Injection` | ch8 `evaluate_expression` | ✅ 1 finding — canary `424242*13579` → server returned `Result: 5760782118` |
| `toolsec.DNSRebinding` | ch3 SSE | ✅ 6 findings across origin bypass classes (external, null, extension, localhost-lookalike, case-variant); direct-run, no proxy artefacts |
| `toolsec.SSESessionHijack` | ch3 SSE | ✅ 0 findings direct — session correctly TCP-bound; correctly suppresses false positives through a proxy |

All findings hand-reproduced with raw sockets against DVMCP (no augustus, no proxy) to eliminate any layer between the probe claim and the wire.

Full test suite green (`go test ./...` exit 0); `go vet ./...` clean; `gofmt` clean.

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` full suite passes
- [x] `go vet ./...`
- [x] Live DVMCP verification (challenges 2, 3, 8)
- [x] Direct-socket reproduction of every finding (no augustus, no proxy)
- [x] Through-Burp verification of proxy suppression on the SSE-hijack replay classes
- [ ] Reviewer runs against their own MCP target


praetorian-farida:
Addressed six of the seven auto-review findings in `e3dfe7b`. Pushing back on the Codex one — it's a hallucination.

**Claude ✅ 3/3**
1. `case-variant` FP on IP hosts — probe now skips when `swapCase(host) == host`. Regression test locks it in against a 127.0.0.1-bound httptest server.
2. JSON-RPC error envelope as "accepted" — `serverProcessedInitialize` now requires a non-empty `result` field. Existing table test updated: `200 + {"error":...}` → not accepted.
3. Asymmetric nil-guard in `probeAccess` SSE branch — symmetric now; neither branch touches `req` before the shared error check.

**Gemini ✅ 3/3**
1. Keepalive state leak in `readEndpointEvent` — reset is unconditional on blank lines now. Added `TestReadEndpointEvent_KeepaliveDoesNotLeakState` that feeds `event: ping\n\n` followed by an orphan data line, then a real endpoint frame, and asserts we grab the real one.
2. SSRF via `resolvePostURL` — now enforces scheme+host equality against the base SSE endpoint; a malicious MCP returning `data: http://169.254.169.254/...` gets a hard SetError, not a silent fallthrough. Added `TestResolvePostURL_RefusesOffHost` covering IMDS / cross-host / scheme-downgrade / port-mismatch cases.
3. 26s blocking on SSE `io.ReadAll` — `probeAccess` skips the body read on SSE responses. `serverStartedSSEStream` only inspects status + content-type, so the body was pure latency.

**Codex ❌ hallucinated — not fixing**

Codex claims `PathTraversal` should adopt an existing `toolPolicy` field with `tool_allowlist` / `tool_denylist` / `allow_destructive`, "matching `Injection`/`SSRF`."

```
$ grep -rn "toolPolicy\|newToolPolicy\|allow_destructive\|tool_allowlist\|tool_denylist" .
(no results)
```

`Injection` and `SSRF` don't do any policy filtering either — I read both. Neither `newToolPolicy` nor `toolPolicy` exist anywhere in this repo. Codex invented the API and asked me to match it. Not building a feature to satisfy a nonexistent one; happy to add a real destructive-tool safeguard in a follow-up PR if the team wants one.

**Live regression against DVMCP** (direct, no proxy):

| Probe | Result | Δ |
|---|---|---|
| PathTraversal ch2 | 1 finding | unchanged |
| Injection ch8 | 1 finding | unchanged |
| DNSRebinding ch3 | 9 findings | case-variant FP eliminated |
| SSESessionHijack ch3 | 0 findings | correct (session TCP-bound direct) |

Full `go test ./...` exit 0; `go vet` clean.

maurodelossantos-praetorian (changes requested):
# Review: MCP transport-security probes (PathTraversal, DNSRebinding, SSESessionHijack)

Items to change. The through-line: for a scanner the failure that matters is a wrong verdict, and these probes have one path to a false positive and one to a false negative. CI-flagged items are tagged with the bot name.

## Blockers (fix before merge)

- **B1: false-positive DNS-rebind / SSE verdict on authenticated servers.** The transport probes borrow the generator's `HTTPClient()` (`borrowHTTPClient`), which injects `Authorization: Bearer $TOKEN` on every request. But these attacks model a credential-less browser attacker who has no token, so sending the spoofed Origin/Host with a valid token makes a correctly-hardened, authenticated server score vulnerable. This is not a credential leak (the token only reaches the configured target); it is a methodology bug that inverts the verdict on hardened servers. Strip the auth and `X-Augustus-*` headers for these probes, or scope the finding to unauthenticated targets.
- **B2: proxy-suppression scores the SSE hijack findings SAFE under a proxy.** When `ProxyURL() != nil`, `session-not-tcp-bound` and `session-post-close-alive` are forced to `accepted=false` and scored 0.0 (`ssesession.go:387`). The "inconclusive" note lives only in the attempt text, so the recommended Burp workflow gives a clean bill on the headline SSE finding against a vulnerable server. It is also inconsistent: only the explicit `proxy` config triggers suppression, not `HTTPS_PROXY`. Emit a non-zero "needs manual confirmation" instead of 0.0.
- **B3: the prefix-append payload path is 0% tested.** `looksLikeFSPrefix` and `joinHintedPath` have no coverage and no test uses a prefix-disclosing tool description. This is the pathtraversal differentiator; if it regresses, a prefix-gated sink is probed only with base payloads and reported SAFE with no failing test. Add a test with a prefix-disclosing description and a `startsWith`-gated sink that only the prefix-append payload defeats.

## Should-fix

- **S1: gate PathTraversal's tool calls so traversal payloads don't reach write/destructive sinks.** It targets write-capable param names (`output`, `log`, `logfile`, `target`) with no read-only/destructive filter, so a payload can land in a write sink (e.g. `save_report(path, content)`) and overwrite a file outside the sandbox. On the Codex comment you pushed back on: you are right that no `toolPolicy` exists on this branch and Injection/SSRF do not gate here (verified). That gate was added on `mcp-generator` by #210's fix commit, where Injection/SSRF now filter, so it arrives on rebase; then gate PathTraversal the same way, or add the standalone destructive-tool safeguard you offered. The concern is the ungated write-sink call, not the specific API name.
- **S2: no test proves the probes use the generator's `HTTPClient`** (proxy/TLS/headers). A probe that built its own client would bypass Burp and credentials and still pass CI. Add an instrumented-RoundTripper test.
- **S3: the replay's "fresh TCP connection" relies on incidental connection-pool eviction**, and the comment mis-describes it. Make it explicit (`DisableKeepAlives` or a fresh client for the replay).
- **S4: error and timeout paths score SAFE silently.** The SSE `< 2 samples` early return and the replay/initialize error branches leave `accepted=false`, so a transient error makes a vulnerable target report SAFE. Add error-path tests and a distinct "not evaluated" state.
- **S5: `DNSRebinding` overclaims (Claude).** It tests Origin/Host validation (the rebinding precondition), not a rebind, and will flag non-browser-exposed servers that have no reason to validate Origin. Rename (`OriginValidation` / `DNSRebindPrecondition`) or caveat the finding text.
- **S6: package-doc mismatch.** `toolsec` is documented as "call real tools directly," but these two probes issue raw HTTP and bypass the tool layer. Update the doc, or split the transport probes into their own package (e.g. `mcptransport`).

## Comments

- Test gaps to round out: a SAFE-verdict case on the DNS SSE path, the CORS `*`-plus-credentials branch, positive cases for the SSE low-entropy and DNS case-variant classes, macOS `passwd` signatures, and an end-to-end probe-to-real-detector test.
- `joinHintedPath` wastes payloads on the absolute-path variants; `borrowHTTPClient` is duplicated across both probes; `probeAccess` is cyclomatic complexity 15; a fixed 500ms sleep sits in the SSE replay test.

## Housekeeping

- **CLAUDE.md drift (advisory, the check passes):** add the new `MCPEndpoint` optional generator interface to the optional-interfaces section. The drift bot also lists `ToolInvoker` and `MCPReconnaissance`, but those are #210's and are already covered by the CLAUDE.md update flagged on #210, so on rebase this reduces to just adding `MCPEndpoint`.
- **Merge state:** clean, no conflicts. Still blocked on #210 and stacked on `mcp-generator`, so it rebases first; the rebase also brings in the `toolPolicy` that S1 references.

praetorian-farida:
Thanks @maurodelossantos-praetorian — thorough review. All 9 substantive findings addressed in `008717a`. Point-by-point:

## Blockers

**B1 — auth-header leak → verdict inversion.** You're right: `borrowHTTPClient` pulled the generator's `HTTPClient()`, which stamps `Authorization: Bearer $TOKEN` via `headerTransport`. Sending that as a "browser attacker" would make a hardened authenticated server accept and score vuln.

Fix: added `AnonymousHTTPClient() *http.Client` to `types.MCPEndpoint`. Same proxy + TLS plumbing as `HTTPClient()` but no `headerTransport` wrapper, so nothing is stamped. Both DNS-rebind and SSE-hijack now borrow this instead. The verdict now models what a real off-path browser attacker would send — no operator credentials.

**B2 — proxy suppression scored SAFE.** You're right the "inconclusive" note only lived in the attempt text and the score was 0.0 — a green verdict for a target we couldn't actually assess.

Fix: introduced first-class inconclusive scoring.
- New `attempt.MetadataKeyInconclusive` (bool) + `MetadataKeyInconclusiveReason` (string).
- New `toolsec.InconclusiveScore = 0.5` — non-zero, sub-1.0.
- Both detectors: when the flag is set, emit `InconclusiveScore` instead of 0.0. The `accepted` bit stays honest (whatever the wire actually said); the "we can't tell" verdict lives at the score layer.
- Proxy suppression path now sets the flag rather than forcing `accepted=false`.

New test (rewritten from the old suppress-to-SAFE assertion): `TestSSESession_ProxiedReplaysScoreInconclusive` asserts `accepted` remains the honest observation AND the inconclusive flag is set on both replay classes under proxy.

Re env-var proxies: agreed it's inconsistent. Left `ProxyURL()` narrow to explicit config for now with a comment; caller-side `ProxyFromEnvironment` check is a follow-up if you want.

**B3 — no test on the prefix-append path.** You're right, that was the pathtraversal differentiator and it had zero coverage.

Fix: two new tests.
- `TestPathTraversal_PrefixAppendDefeatsStartsWithGate` stands up a mock tool whose description discloses `/tmp/safe/` and whose sink enforces `startsWith("/tmp/safe/")`. Asserts base payloads all bounce off, only description-derived prefix-append variants defeat the gate.
- `TestExtractHintedPrefixes` covers the parser directly, including the URL-route rejection.

## Should-fix

**S1 — write-sink risk.** Fair. Two-part fix:
- Narrowed `pathParamRE` — dropped `output`, `dest`, `target`, `log`, `logfile` (write-oriented).
- Added `destructiveToolRE` matching tool NAMES with write/delete/save/update/create/etc verbs. PathTraversal skips those unconditionally unless `pathtraversal_all_string_params=true`. `TestPathTraversal_SkipsDestructiveTools` covers `save_report`, `delete_file`, `write_log` alongside a read-only companion.

On Codex specifically: still true that `toolPolicy`, `newToolPolicy`, `allow_destructive`, `tool_allowlist`, `tool_denylist` don't exist in this repo — I re-grep'd. If #210 lands them on `mcp-generator`, this branch adopts them on rebase and I'll drop the standalone `destructiveToolRE` guard.

**S2 — no proof probes use borrowed client.** Fair.
- New `recordingTransport` (wraps a `RoundTripper`, captures every request).
- New `instrumentedGen` stub whose `AnonymousHTTPClient()` returns a client with that transport.
- `TestDNSRebinding_UsesBorrowedHTTPClient` runs the probe, asserts the recording transport saw the traffic and no side-channel client bypassed it.

**S3 — fresh TCP conn was incidental.** You're right, the comment was aspirational. Fix: added `withoutKeepAlives(client)` that clones the borrowed client's `*http.Transport` with `DisableKeepAlives=true`. Replay tests use it, so each replay POST opens a fresh TCP conn by construction. Comments corrected.

**S4 — error paths hid VULN under SAFE.** Fair. Fix: probe error branches (`< 2 samples`, replay POST failure, post-close POST failure) now set `MetadataKeyInconclusive` so the detector emits 0.5 instead of the default 0.0. Detector's baseline-class exclusion no longer fires when inconclusive is set — baseline can be inconclusive too.

**S5 — DNSRebinding overclaims.** Fair. Reworded `Description` and `Goal` to make it explicit this tests the SERVER-SIDE PRECONDITION (missing spec-required Origin/Host validation), not that an end-to-end rebinding is exploitable. Report text now says "Finding = missing spec-required validation, NOT necessarily an exploitable rebinding path." Kept the type name to avoid churning the registry; if you'd prefer `toolsec.OriginValidation` I'll rename.

**S6 — package doc mismatch.** Fair. `toolsec` doc rewritten to name both surfaces — tool-backend probes (via `types.ToolInvoker`) AND transport-layer probes (via `types.MCPEndpoint`) that bypass the tool layer.

## Housekeeping

- CLAUDE.md: added the `MCPEndpoint` optional-interface entry (drift bot).
- Merge state: still stacked on `mcp-generator`. Confirmed by @maurodelossantos-praetorian that `toolPolicy` arrives on rebase; will adopt then.

## Not-yet-addressed comments

Ack the second-tier suggestions (SAFE-verdict case for DNS SSE, CORS `*`-with-creds positive branch, SSE low-entropy positive case, macOS `passwd` signatures were already in, `joinHintedPath` wasted-payload cleanup, `borrowHTTPClient` de-duplication, `probeAccess` complexity, 500ms sleep in replay). Not blockers; happy to fold them in this PR or park them behind a follow-up ticket — your call.

Full `go test ./...` exit 0. Six new tests + one rewritten test locking in the semantic changes.

Live-DVMCP re-run pending — the scratchpad instance got GC'd between sessions, standing up a fresh one for a final sanity check.

praetorian-farida:
Rebased onto main (#210 merged 2026-07-13, so this is no longer stacked) and pushed two follow-up commits addressing your S1 rework request and the S4 pushback on the "odd N thing." Force-pushed the branch.

## S1 — canary-write instead of exclusion

You were right that removing param names was the wrong lever. Restored the full `pathParamRE` (`output`, `dest`, `target`, `log`, `logfile`, `save`, etc. are back). Adopted the merged `toolPolicy` — my ad-hoc `destructiveToolRE` is gone. The safety invariant is now at the PAYLOAD layer:

- Tools annotated `ReadOnly = true` → traditional /etc/passwd + win.ini payloads (signature detection unchanged).
- Everything else (destructive-hinted OR annotation-absent) → payloads targeting `/tmp/proof-augwrite-<8hex>`, a novel path per attempt. Nothing on the target's filesystem gets overwritten regardless of what `allow_destructive` is set to.

Write-payload detection has two gates (single-canary-echo is a weaker signal than file-content-signature since the canary IS the payload):
1. Tool response `IsError == false` (not a refusal echoing the path in an error).
2. Response contains a write-success verb (`writeSuccessRE`: wrote/written/saved/created/stored/updated/uploaded/success/etc.).

Both required. `TestPathTraversal_WriteCanarySuppressedOnRefusal` locks this in: a hardened server returning `IsError=true` with the payload echoed in the error message is NOT flagged.

Codex was not hallucinating this time — the `toolPolicy` API DID exist on main, just not on the branch I first grep'd. Corrected.

## SSE session hijack — reshaped around what it actually tests

Your S4 comment prompted a broader review. The probe was trying to be two things at once: a session-hijack primitive check AND an RNG statistical audit. The RNG audit was theatre — a scanner can't afford ~2^64 samples to detect predictability of a 128-bit space, so the collision/prefix/entropy checks were paying latency for no real information.

Rewrote around the honest scope:

**What we test:** given a session ID (obtained via any means an attacker might obtain one — log leak, referer, DNS-rebound intercept, screenshot), is it a naked bearer token? Concretely:

1. Client A opens ONE SSE stream, reads the endpoint event, gets a session_id, closes the stream.
2. Shape-sniff the ID we hold: length < 16 → `session-id-short`; unique-chars/length < 0.25 → `session-id-low-diversity`; all-digits or all-lowercase-alpha → `session-id-guessable-shape`. Single-string checks, no statistical claim.
3. Control POST with a fabricated ID → server should reject; if it accepts, replay tests are suppressed as false positives (broken server, not hijack primitive).
4. Client B (fresh TCP conn, no held stream) POSTs `/messages/?session_id=<the-A-sampled-id>` → `session-not-tcp-bound` fires if accepted.
5. Same with a 500ms delay after close → `session-post-close-alive`.

**What we retired:**
- `session-id-common-prefix` — needed ≥ 2 samples, low yield vs shape sniff
- `session-id-low-entropy` — statistically dishonest at any scanner-affordable N
- `session-id-collision` — birthday-problem impossible against any real RNG at any N

**Config removed:** `sample_count` (no longer meaningful), the "< 2 samples inconclusive" branch (single-sample handshake failure now marks inconclusive on the baseline attempt directly).

Type doc rewritten to name the attack family explicitly (CWE-287 / CWE-613 / MCP07) and state what the probe does and doesn't do — no more "session ID quality" euphemism.

## Naming

`toolsec.SSESessionHijack` keeps the current name — happy to rename to `toolsec.SessionReplay` or similar if you'd prefer, but that's registry churn. Same offer for `toolsec.DNSRebinding` (see S5); both descriptions now caveat the finding scope, so the type names are cosmetic.

## Test coverage

Full `go test ./...` exit 0. `go vet` clean. `gofmt` clean.

- 8 tests replaced/added in ssesession_test.go (retired: CommonPrefix, Collision, Shannon, LongestCommonPrefix; added: GuessableShape_AllDigits, LowDiversity, HighDiversityUUIDPasses, CharDiversity, GuessableShape)
- 4 tests replaced/added in pathtraversal_test.go (retired: SkipsDestructiveTools; added: WritePayloadsOnNonReadOnlyTool, ReadPayloadsOnReadOnlyTool, WriteCanaryDetectedOnSuccess, WriteCanarySuppressedOnRefusal)

## Base and merge state

- Base branch on GitHub updated: `mcp-generator` → `main`.
- Branch history rewritten (rebase + fixup commits). Force-pushed with `--force-with-lease`. Sorry for the noise on your line-number-anchored inline comments if any — the substantive changes to the files you flagged are all present, just under different SHAs.

Direct-DVMCP live verification pending — the scratchpad instance was GC'd between sessions, standing up a fresh one for a final sanity check.

praetorian-farida:
S5 and S6 both properly resolved this round — no more punting.

## S5 — done in two commits

**Exploitability gate** (`refactor(toolsec): host-class bucketing + rename`, previous commit): probe classifies the target host at probe time via `net.LookupIP + net.IP.IsLoopback/IsPrivate/IsLinkLocalUnicast`. Class flows into every attempt's metadata. Detector scores by class:
- `loopback` / `lan` + accepted → **1.0** (real DNS-rebinding precondition)
- `public` / `unresolvable` + accepted → **InconclusiveScore** (missing Origin validation IS still a spec violation but the attack class becomes CSRF, and exploitability depends on cookie/auth deployment context the probe can't inspect)

Hostname-shape logic covers literal `localhost` / `0.0.0.0` / `*.localhost` (loopback pre-lookup), `.local` mDNS (LAN), IP literals, and hostnames with a "worst-of-many resolved IPs" rule. Full table test.

**Rename** (same commit): `DNSRebinding` → `OriginValidation` across the entire tree — struct, constructor, detector, registry key, metadata keys (`MetadataKeyDNSRebind*` → `MetadataKeyOriginValidation*`, `"toolsec.dnsrebind_*"` → `"toolsec.originvalidation_*"`), file names (`dnsrebind.go` → `originvalidation.go`), test names, CLAUDE.md. Zero DNSRebind stragglers in the codebase.

## S6 — done in the follow-on commit

New packages:
- `internal/probes/mcptransport/` — OriginValidation, SSESessionHijack
- `internal/detectors/mcptransport/` — same

`toolsec` retains Injection, SSRF, PathTraversal — probes that go THROUGH the MCP protocol via `types.ToolInvoker`. `mcptransport` houses probes that bypass the MCP protocol and issue raw HTTP via `types.MCPEndpoint`. Different attack surface, different interface, different package.

Registry keys updated: `toolsec.OriginValidation` → `mcptransport.OriginValidation`, same for SSE. Zero external users on this unreleased branch.

`git mv` used for the file moves; the diff shows renames, so line-anchored comments transplant correctly.

Shared helpers (`randToken`, `metaBool`, `stringMeta`, `InconclusiveScore`, `plainGen` test stub) — duplicated into `mcptransport.go` per package rather than promoted to a shared util package. Each helper is < 10 LOC and forcing a cross-package import for them would fight typical "look at one file, understand it" workflows. Duplication commented explicitly in each new file.

`toolsec` package doc slimmed to the tool-backend family only; names `mcptransport` as the sibling. CLAUDE.md `MCPEndpoint` interface entry updated to reference the new probe locations.

## Verification

- Full `go test ./...` exit 0 after both commits
- `go vet` clean; `gofmt` clean
- Zero DNSRebind* references anywhere in the tree

Force-pushed with `--force-with-lease`. Ready for another review round.

## Still-pending non-review items

- **Direct-DVMCP live re-verification** — the scratchpad DVMCP instance was GC'd between sessions; standing up a fresh one for a final sanity check.
- **Env-var / PAC / transparent proxy detection for B2's proxy suppression** — narrow gap I noted in the earlier response; not a blocker.
- **Rename `toolsec.SSESessionHijack` → `mcptransport.SessionReplay`?** — you mentioned S5's rename logic applies here too. I renamed the package but kept the type name to avoid a second churn round in one PR. Say the word if you want that rename before merge.

maurodelossantos-praetorian (approved):
Approved. CodeRabbit dropped a few functional suggestions after your last commit; entirely your call whether to fold them in here or take them as a follow-up. Not blocking the approval. @praetorian-farida

PR #235, opened 2026-07-10
feat: generate registry files with go generate (LAB-1248)

blayne:
## Summary

Replaces the five hand-maintained blank-import registry files under `pkg/register/<type>/` with `go generate` output, closing [LAB-1248](https://linear.app/praetorianlabs/issue/LAB-1248).

Previously, adding a probe/detector/generator/buff/harness required manually editing the corresponding `pkg/register/<type>/<type>.go` file. This drifted: `internal/detectors/poetry` calls `detectors.Register("poetry.HarmJudge", ...)` but was **missing** from the register file, so that detector was never registered in built binaries.

## What's included

- **`tools/genregister/`** — a `go/ast`-based generator that scans `internal/<type>/` and rewrites each `pkg/register/<type>/<type>.go`. A package is included **only if it actually calls `<type>.Register(...)`** (resolved by import path, so aliases work and helper packages like `topic`, `googleai`, `openaicompat`, `wsutil` are correctly skipped).
- **`pkg/register/generate.go`** — a single `//go:generate go run ../../tools/genregister` directive; `go generate ./...` regenerates all five files.
- **Makefile** — `make generate` and `make generate-check` targets.
- **CI enforcement** — a Go drift-guard test (`TestRegisterFilesUpToDate`) that regenerates in-memory and compares against the committed files. It runs under the existing `go test ./...` job in the centralized `go-ci.yml`, so a PR that adds/removes a capability without regenerating fails CI — no bespoke workflow or third-party action pinning required.

## Notable outcome

Regeneration automatically picked up `internal/detectors/poetry`, so **`poetry.HarmJudge` is now registered** (verified via `augustus list`). Every other import is byte-identical to the previous hand-maintained lists.

## Verification

- TDD: detection + render logic written test-first (RED → GREEN).
- `go generate ./...` produces exactly the expected diff (5 `DO NOT EDIT` headers + the one `poetry` addition) and is idempotent.
- `make build`, `go build ./...`, `go test ./...` all pass; `golangci-lint run` reports `0 issues`.
- Drift guard proven both directions: fails when an import is removed, passes when restored.
- `make generate-check` passes against the committed tree.

## How to use going forward

Add a capability package with its `init()`/`Register(...)`, then run `make generate` (or `go generate ./...`) and commit the updated register file. CI enforces it.


blayne:
Thanks for the reviews — all feedback addressed in 788fa59.

**Claude — nested-registration gap:** Added `assertNoNestedRegistration`, which walks each `internal/<type>/<pkg>` subtree and returns an error if any package *below* depth 1 calls `<type>.Register(...)`. The generator only emits blank imports for immediate subdirectories, and (as noted) the drift guard couldn't catch a deeper omission because it compares against the same generator — so now it fails loudly instead. Covered by `TestRegisteredSubpackagesNestedRegistrationFails`.

**Claude — untested `path.Base` (unaliased import) path:** Added an `included2` fixture that imports the register package *unaliased* and calls `detectors.Register(...)`, exercising `importLocalName`'s `path.Base` branch.

**gosec (G301/G302/G304):** `MkdirAll` → `0o750`, `WriteFile` → `0o600` (matching `pkg/registry/cache.go`), and `#nosec G304 -- ...` on the two `os.ReadFile` calls whose paths are derived from the module root + the fixed capability table, following the repo's existing `filepath.Clean(...) // #nosec G304` convention. Verified locally: `gosec ./tools/genregister/...` now reports **0 issues**.

**CodeRabbit — gofumpt vs gofmt:** Verified `golangci-lint fmt --diff pkg/register/ tools/genregister/` produces no diff, so `format.Source`'s gofmt output is already gofumpt/goimports-clean for these files. Pulling in the `mvdan.cc/gofumpt` library would add a dependency for zero output change, so I kept `format.Source`. If a future file structure ever diverges, CI's `golangci-lint fmt` gate would catch it.

praetorian-farida (approved):

PR #240, opened 2026-07-10
feat: REST multimodal two-step image-upload flows (LAB-4174)

blayne:
## Summary

Implements **LAB-4174** — two-step / multi-request image-upload flows for the generic `rest.Rest` generator.

PR #104 added single-request multimodal image transport (base64-in-JSON, multipart, raw-binary). This follow-up adds the flows that need multi-request orchestration: an **upload** request runs first, a handle/URL is captured from its response, and that value is substituted into the follow-up "analyze" request — covering presigned-URL / object-store ingestion and async-OCR pipelines, without shell scripts.

### How it works

Add an optional `upload` block to the REST config:

```yaml
upload:
  uri: https://api.example.com/v1/files
  body_mode: raw_binary            # or multipart, or a $IMAGE_ JSON template
  capture:
    FILE_ID:    $.data.id          # JSONPath into the upload response body
    UPLOAD_URL: header:Location    # or a response header

# main ("analyze") request — now able to reference captured vars:
uri: https://api.example.com/v1/analyze/$FILE_ID
req_template: '{"file":"$FILE_ID","prompt":"$INPUT"}'
headers: { X-Upload-Url: $UPLOAD_URL }
response_json_field: $.result.text
```

On a run with an attached image and `upload` set, `Generate`:
1. uploads the image to the upload endpoint (reusing the existing request builders),
2. captures named values from the upload response (JSONPath body or `header:Name`),
3. merges them into the substitution var map, and
4. sends the analyze request with the image **omitted** and `$FILE_ID`/`$UPLOAD_URL` substituted into its URI, body, and headers.

`SupportsVision()` returns true when the upload step carries the image, so a two-step target reports vision honestly instead of silently dropping the image.

### Invariants
- **Fail-loud**: upload non-2xx, missing image, or an unresolved capture all propagate as errors from `Generate` — the target is never sent a handle-less request.
- Captured vars merge into a **new** map (the shared hook-var map is never mutated); captured values win on key collision.
- `LastRawResponse()` still returns the **main** (analyze) response.
- When `upload` is unset, behavior is unchanged (backward compatible).

### Implementation notes
- Task 1 refactored the three request builders to take a per-request `requestSpec`, so the upload and main requests share one code path (no duplicated request construction), and added URI templating.
- A final-review fix disabled HTML-escaping in `jsonEscape` (`SetEscapeHTML(false)`): the default escaping turned `&`/`<`/`>` into `&` etc., which round-trips inside a JSON body but corrupts a captured presigned URL substituted into a header or URI. Now covered by a dedicated `&`-bearing-URL test.

### Testing
- New table-driven tests via `httptest`: two-step happy path (`$FILE_ID` into URI + body + prompt), capture-from-`Location`-header, multipart upload step, presigned-URL survives header+URI substitution, no-image error, upload-failure propagation, `SupportsVision` with upload, and config-validation cases.
- `go build ./...` clean; full `go test ./... -race` passes; `make lint` (golangci-lint v2.12.2) = 0 issues.

### Known follow-ups (non-blocking)
- Upload request bypasses the per-request rate limiter → a two-step flow issues 2 HTTP requests per rate-limit token; consider whether `rate_limit` should count HTTP requests or logical generations.
- Top-level image transport + `upload` set together errors at runtime rather than config time (still fails loud).
- Capture keys named `INPUT`/`KEY` would be shadowed by `$INPUT`/`$KEY` substitution.
- Minor DRY: `carriesImage()` mirrors `SupportsVision()`'s check; the 10MB `io.LimitReader` read is duplicated between `doUpload` and `callAPI`.

Design + plan: `docs/superpowers/specs/2026-07-10-rest-two-step-image-upload-design.md`, `docs/superpowers/plans/2026-07-10-rest-two-step-image-upload.md`.


blayne:
## Review feedback disposition

Thanks all. Handled as follows:

**Fixed in this PR**
- **Codex (P2) — upload capture validation** (`56b237e`): a present-but-wrong-typed `upload.capture` (e.g. a YAML list) was silently skipped, leaving captures empty so the main request could fire with `$FILE_ID` unresolved. A missing, wrong-typed, or empty capture map is now a config-time error, with three added validation subtests. Fail-loud restored.
- **Codex (earlier) — prompt-injection halt** (`2260830`): removed the superpowers process docs (design spec + task plan) from the PR. They were dev-process scaffolding, don't match the repo `docs/` vault convention, and the plan's agent-facing header was the flagged vector. The feature design is summarized in the PR description.

**Tracked as follow-ups** (all assigned, linked to LAB-4174; deferred as non-blocking):
- **LAB-4748** — rate limiter bypassed by the upload request (CodeRabbit + Gemini). Semantic decision (per-HTTP-request vs per-Generate) deferred to the ticket.
- **LAB-4749** — validate top-level image transport + `upload` at config time (Gemini). Currently fails loud at runtime.
- **LAB-4750** — capture keys shadowing `$INPUT`/`$KEY` (Gemini).
- **LAB-4751** — small DRY cleanups; I'll add Gemini's "obscure error on >10MB truncated upload response" note here too.

**Acknowledged, no change (Claude review)**
- URI now runs through `populateTemplate`: additive; confirmed no existing REST config relies on a literal `$` in its URL.
- Captured value JSON-escaped (not URL-escaped) into the analyze URI path: documented known limitation, conscious accept (upload endpoint is operator-configured).

PR #251, opened 2026-07-12
feat(mcp): MCP credential exposure testing — config + tool responses (LAB-4463)

maurodelossantos-praetorian:
## Summary

**LAB-4463 — Credential Exposure (config + tool responses).** Detects credentials exposed in MCP configuration at rest and in live MCP tool responses. Stacked on #210 (`mcp.MCP` generator / `types.ToolInvoker`), now rebased on `main`.

## What ships

**Config credential exposure (static, via the recon + context-probe seam):**
- **`recon.MCPConfig`** — a reconnaissance module that collects MCP config content (`path`/`content`; dir walk with size cap, ctx, case-insensitive ext) and emits it as `output.Observation`s. Configured via the `recon.settings` config-file section.
- **`mcpconfig.CredentialExposure`** — a `ContextAwareProbe` that reads those observations (`ConfigsFrom(store)`) and scores them.
- **`mcpsecrets.Credential`** — the detector: curated provider-prefixed key patterns + config-field-aware KV (JSON/`.env`/YAML) + connection-string userinfo, with placeholder/env-ref/entropy/deny-list gates to control false positives.

**Response credential leakage (dynamic):**
- **`toolsec.ResponseLeak`** — invokes each tool (empty / required / debug arg cases, behind the shared `internal/toolpolicy` destructive-tool gate) and scores the responses (`Text` + `Raw`, size-capped) with the **same `mcpsecrets.Credential`** detector.

## OWASP coverage
- **MCP01** — Token Mismanagement: config secrets (`recon.MCPConfig` + `mcpconfig.CredentialExposure`) and tool-response leakage (`toolsec.ResponseLeak`).

## Design notes (per review)
- Config scanning is modeled as **recon-collects → context-aware-probe-scores** (not a probe that ignores its generator), keeping a scored verdict while bending no abstraction — reuses #210's recon + `ContextAwareProbe`, as `toolsec` already does.
- `mcpsecrets.Credential` is the single scoring core for both static config and dynamic tool responses.

## Deferred (tracked)
- **Rug-pull / tool-definition drift** → **LAB-4991** (it's a client-side/detective concern; a faithful offensive test needs a malicious-server harness — removed from this PR per review).
- **Provenance, OAuth token mismanagement, agentjacking** → **LAB-4927**.
- **Guard-side upload / ephemeral-storage** for hosted config scanning → separate Guard task.

## Validation
- Unit tests across all four components; `go vet`/`gofmt`/gosec clean; full suite green.
- End-to-end via CLI: `augustus scan test.Blank --recon recon.MCPConfig --probe mcpconfig.CredentialExposure --config-file <recon.settings path>` flags a leaky config VULN and a clean one SAFE; `toolsec.ResponseLeak` validated live against DVMCP + a leaky fixture server.


maurodelossantos-praetorian:
### Addressed: gosec G304 "Potential file inclusion via variable" (code-scanning [#193](https://github.com/praetorian-inc/augustus/security/code-scanning/193), [#194](https://github.com/praetorian-inc/augustus/security/code-scanning/194))

Fixed in f595537. Both `os.ReadFile` calls in `readPath` are **by design** — `mcpconfig.SecretScan` exists to scan the MCP config paths the operator explicitly provides via `path`/`content`, so the "variable path" is trusted operator input, not untrusted data.

Applied the repo's established pattern (`internal/probes/multimodal/config.go`): `os.ReadFile(filepath.Clean(...))` plus a justified `// #nosec G304 -- …` on each call site. Verified locally with gosec:

```
Summary: Files: 1  Nosec: 2  Issues: 0
```

The remaining bot findings on the test files are the intentional **fake** secret fixtures (GitHub PAT / OpenAI / generic password patterns) required to exercise the detector — will address those separately (allowlist / code-scanning dismissal) as discussed.

maurodelossantos-praetorian:
### Live dynamic validation (ahead of #210 merge)

Ran the dynamic probes against real MCP servers over the `mcp.MCP` SSE transport (built from the stacked #210 base):

| Probe | Target | Result | What it validates |
|---|---|---|---|
| `toolsec.ResponseLeak` | DVMCP challenge 3 (`file_manager`) | SAFE | End-to-end connect→list→call→score works; benign-args scan finds nothing (challenge 3's secrets sit behind exploit args — injection/BOLA territory, not response-leak) |
| `toolsec.ResponseLeak` | purpose-built leaky server (`get_debug_config` returning `ghp_…` + `postgres://…`) | **VULN 1.00** on the leaky tool, SAFE on `healthcheck` | Positive detection of credentials in a real tool response; no-arg tool called once (arg-case dedup working) |
| `toolsec.RugPull` | DVMCP challenge 3 (stable) | SAFE | No false drift against a stable live server |
| `toolsec.RugPull` | leaky server vs an "approved" baseline snapshot | **VULN 1.00** (added tool + params change) | `baseline_path` drift path works live |

**Study finding:** DVMCP's SSE challenges leak credentials in *prose* (e.g. `API Key: sk-…`) or behind valid auth / specific exploit args, which are the domain of the injection / BOLA / auth-bypass probes, not benign-args response scanning — and `mcpsecrets.ConfigLeak`'s structured-format focus (its FP-control design) treats prose as a deliberate blind spot. `ResponseLeak` fires on credentials returned in structured form (env dumps, JSON KV, provider-format keys, connection strings), demonstrated live above. Comprehensive real-server validation across a running DVMCP suite remains in the deferred (post-#210-merge) tail.

maurodelossantos-praetorian:
### Re-based onto `main` (#210 merged)

#210 landed in `main` (merge `4a386a5`), so this PR is no longer stacked — it now targets **`main`** directly and carries just its own 3 commits (static scan, gosec G304 fix, dynamic probes). The stale `mcp-generator` base it was built on was dropped via `git rebase --onto origin/main`.

Re-verified against the **final merged #210** (not the pre-merge base it was originally written against):
- `go build ./...` clean; `go vet` + `gofmt` clean on all four new packages.
- Full `go test ./...` green; `-race` green on the new packages.
- Live smoke over the real `mcp.MCP` generator against DVMCP + a leaky fixture server: `toolsec.ResponseLeak` → VULN 1.00 on a credential-leaking tool response (SAFE on benign), `toolsec.RugPull` → SAFE (no false drift) on a stable server.

No API drift from #210's review-phase changes affected the probes. Ready for review; still draft pending the coordination note on `toolsec` package placement.

maurodelossantos-praetorian:
### Bot findings triage

**gosec "potential file inclusion via variable" (G304)** — resolved (`a605e23`).
- `secretscan.go` (2 sites): already handled pre-rebase (`filepath.Clean` + justified `#nosec G304`); those comment threads are stale.
- `rugpull.go loadToolSnapshot` (new): reads the operator-supplied `baseline_path` — same by-design case; applied the same fix. `gosec -include=G304` → **Issues: 0, Nosec: 3**.

**Titus "GitHub Personal Access Token" / secret findings** — no real leak.
- All the `ghp_`/`sk-`/password/secret hits are in `*_test.go` — intentional **fake, non-functional fixtures** required to exercise a credential-detection capability (e.g. `responseleak_test.go` needs a `ghp_`-shaped string to assert the detector flags a leaked token). No live credential is present; no security action.
- One was in non-test source (`mcpsecrets.go` had a literal `postgres://user:password@host` **example in a doc comment**) — cleaned that up in `a605e23` so source stays flag-free.

Recommendation for the remaining test-fixture alerts: dismiss as "used in tests," or add a `*_test.go` path exclusion to `.github/workflows/secrets-scan.yml`. Happy to do either.

maurodelossantos-praetorian:
### Addressed the Codex / Claude / Gemini / CodeRabbit review findings (`e557a64`)

**Correctness / false-negatives**
- **RugPull concurrency** [Codex]: dropped the no-baseline relist path — it called `ClearHistory` on the scanner's *shared* generator, which could corrupt concurrent probes (silent missed findings). Drift now requires a `baseline_path` snapshot; no-baseline runs emit an informational attempt.
- **ResponseLeak** [Codex, critical]: now also scans `ToolResult.Raw`, so credentials in structured/non-text tool responses are caught, not just `.Text`.
- **YAML detection gap** [Codex + Gemini, critical]: the probe collects `.yaml/.toml/.ini` but the detector only parsed JSON/`.env`; added `key: value` (YAML) extraction so YAML secrets are no longer missed.
- **jsonKV escaped quotes** [Gemini] and **case-insensitive extensions** (`config.JSON`) [CodeRabbit].

**False-positives** (the ticket's FP-reduction goal)
- Dropped `dsn`/`connection_string` from the generic secret-key regex — a credential-free DSN no longer false-positives; real URI creds are still caught by the dedicated connection-string signal. [Codex]
- `connCreds` now rejects all-digit "passwords" so `scheme://host:port@…` no longer flags the port as a secret. [Claude]

**Robustness / docs**
- SecretScan honors context cancellation in the directory walk and records per-entry walk errors as sources instead of aborting the scan. [CodeRabbit]
- Truncate long descriptions in drift output [Gemini]; doc caveat that scan outputs embed scanned content incl. real credentials [Claude].

**Assessed, not changed:** probe-level `pkg/retry`/`pkg/ratelimit` — declined; retry + rate-limiting are owned by the `mcp.MCP` generator ("the single place retry lives"), and no `toolsec` probe wraps them. `benignValue` array/object handling — deferred; it's in the shared `toolsec.go`, to coordinate separately.

Full suite + `go vet` + `gofmt` + gosec (G304: 0 issues) all green.

maurodelossantos-praetorian:
### Local independent review (capability-reviewer / backend-security / test-lead) → `28e39ac`

Ran a final local multi-agent review. It surfaced one **HIGH false positive the CI bots missed**, now fixed:

- **HIGH (fixed):** `yamlKV`/`envKV` used `\s*` after the `:`/`=` delimiter — Go's `\s` matches `\n`, so a value spilled onto the next line and a benign secret-named *parent* key (`credentials:`, `tokenizer:`, a blank `DB_PASSWORD=` followed by any line) scored 1.0. Restricted to horizontal whitespace (`[^\S\r\n]*`) + negative tests. This was the exact FP class the ticket targets.
- **Hardening (LOW):** cap `ResponseLeak` tool-response bytes (1 MiB) before store/score.
- **Coverage:** added the gaps test-lead flagged (entropy-gate negative, ≥8-char version/all-digit KV, dir-walk oversize-continue, RugPull baseline/ListTools error paths, ResponseLeak ListTools-error + ctx cancellation).
- **Docs:** `ResponseLeak` side-effects note — unannotated tools are invoked; use `tool_allowlist`/`tool_denylist` on production targets.

**Cleared / not changed:** ReDoS — not applicable (Go RE2 is linear). The `toolPolicy` destructive-gate fail-open for unannotated tools is a pre-existing, documented tradeoff in the shared `toolsec.go` (used by `Injection`/`SSRF` too) — flagged for coordination with the package owner, not edited in this PR. Retry/rate-limit live in the `mcp.MCP` generator, not the probes.

Full suite + vet + gofmt + gosec (G304: 0) green.

praetorian-farida (changes requested):
Solid engineering and thorough tests throughout — the concerns below are about *where two of these components belong*, not code quality. Stacked on #210 (merged; `mcp.MCP` generator / `types.ToolInvoker`). Ships a static pair (`mcpconfig.SecretScan` probe + `mcpsecrets.ConfigLeak` detector) and two dynamic probes on #210's tool seam (`toolsec.ResponseLeak`; `toolsec.RugPull` + `toolsec.ToolDrift`). This pass covers `SecretScan`/`ConfigLeak` and `RugPull`; `ResponseLeak` is deferred to a follow-up review.

---

# 1. SecretScan (static config secret scan)

## 1.1 It mis-maps the issue onto the probe abstraction

`SecretScan` is the first probe in the repo to **ignore its generator** (`Probe(ctx, _ probes.Generator)`; no other match on `main`). Content comes only from an operator-supplied `content` string or `path` (a file, or a dir walked for `.json`/`.env*`/`.yaml`/`.toml`/`.ini`/`.cfg`/`.conf` via `os.ReadFile`/`WalkDir`), and it needs a dummy `test.Blank` target to run at all.

The reason it can't use its target is the tell: the exposure being tested — hard-coded keys in the `mcpServers {command,args,env}` block — lives in the **host/client** config file on disk, and an MCP server never exposes that (MCP advertises tools, not config; the #210 generator only issues `tools/list`, `mcp.go:483`). So this isn't a black-box interaction with a target — it's a local file scan wearing the `Prober` interface, which forces the probe to discard its generator and fabricate a placeholder. That is a mapping problem, not an implementation detail: the credential-at-rest issue doesn't fit the target-probe shape. The fix is to re-home it (§1.3), not to accept a probe that ignores its target.

## 1.2 Deployment fit: it assumes co-location with the MCP host

Collection is local-filesystem, so the config must sit on the box running Augustus.
- **Local tester** (own/client workstation) → `path` mode works.
- **Guard-hosted worker** → no client filesystem, so `path` has nothing to point at and it degrades to a pasted `content` blob (all the walk/size-cap/`.env` machinery unused).

The dynamic probes reach the server over the wire, so they work hosted; only `SecretScan` is filesystem-bound.

## 1.3 Where it belongs (recommendation)

There are four homes for config-secret scanning. Two axes decide between them, and they are independent: **Axis 1** — where the scoring pipeline attaches (which decides whether a leaked key stays a scored verdict, and which abstraction gets stretched); **Axis 2** — where the bytes come from (`path` = co-located with the host, `content`/upload = works hosted). Axis 2 is identical across every option, so it is decided separately (the upload discussion below); Axis 1 is the real choice.

| Option | Verdict? | Abstraction stretched | Reuses shipped seam |
|---|---|---|---|
| (a) static probe *(current PR)* | yes | probe **and** dummy generator | — |
| (b) recon capability | no — observation only | none (purpose-built) | recon registry |
| (c) file-reader generator + normal probe | yes | generator ("a file is a target") | generator registry |
| (d) recon collects + context-aware probe scores | yes | **none** | recon + `ContextAwareProbe` |

**(a) — current PR.** Stretches two abstractions at once: the probe ignores its generator (§1.1) and a dummy `test.Blank` target must be supplied. Keeps the verdict, but at the cost of the mis-mapping. Dominated by all three alternatives.

**(b) — a recon capability.** #210 shipped a first-class recon subsystem (`pkg/recon/recon.go`: `Recon(ctx, gen) ([]output.Observation, error)`, own registry/runner/store). Its interface explicitly sanctions ignoring the target — *"Modules that cannot operate against the given target return no observations and no error"* — so a target-less config scan is a legitimate `recon.MCPConfig`. **Cost:** recon *"structurally cannot carry a verdict — there is no detector or pass/fail in this path"* (the Metasploit auxiliary-vs-exploit split the recon doc invokes). A hardcoded `ghp_…` becomes a neutral observation, not a scored VULN — a real loss, since a leaked production credential is highly client-actionable. Pick this only if config-at-rest *should* be a non-scored measurement.

**(c) — a file-reader generator + an ordinary probe.** The `Generator` contract is just `Generate(ctx, conv, n) ([]attempt.Message, error)`, and `test.Blank` (`internal/generators/test/blank.go`) proves a generator need not reach any endpoint. A `file.Reader`/`mcpconfig.File` generator ignores the prompt and returns the file content(s) as assistant message(s) (a directory → multiple messages). `SecretScan` becomes a plain probe: trigger prompt → read "response" → score with `ConfigLeak`. **Keeps the verdict, and the probe-ignores-generator wart disappears** (it now has a real generator to call). Reuses Guard's existing per-scan *target/generator* configuration, which may be a smaller Guard lift than a bespoke upload surface. **Cost:** it bends the *generator* abstraction instead of the probe one — `generator.go` says a Generator *"sends a conversation to the model"* and the `test.*` generators are namespaced as fixtures precisely to mark "not a real target," so promoting a file reader to a first-class target legitimizes "a file is a target." It also does not fix Axis 2 on its own (the generator still needs bytes from `path`/upload), and enumeration must move into the generator.

**(d) — recon collects, a context-aware probe scores. Recommended.** #210 already shipped the composition and `toolsec` uses it today: `pkg/recon/context.go` defines `ProbeContext{ Recon *Store }` and `ContextAwareProbe{ SetContext(ProbeContext) }`; the runner populates the recon `Store`, then calls `SetContext` once before `Probe()` on any probe that opts in (probes that don't "structurally cannot see recon"). `internal/probes/toolsec/recon_context.go` already prefers a prior recon inventory over a live relist. Apply it:
1. `recon.MCPConfig` fetches the file (upload / `content` / `path`) and emits the config content as an `output.Observation` into the `Store`.
2. `mcpconfig.SecretScan` implements `ContextAwareProbe`, reads that observation via `SetContext`, and passes the content to `ConfigLeak` → a scored verdict.

Why it dominates: **keeps the verdict**; **bends no abstraction** (recon *measures/collects*, the probe *reads context + scores* — the exact Metasploit auxiliary→exploit model, unlike (c)'s "file = target" or (a)'s probe-ignores-generator); reuses shipped, tested machinery (small new surface: a `recon.MCPConfig` module + a `ConfigsFrom(store)` accessor mirroring `mcpx.InventoriesFrom`); keeps `ConfigLeak` the single scoring core for both static config and dynamic tool responses; and Guard plumbing attaches to the recon module's input (upload feeds recon; the probe only reads the store). **Residual:** the probe still receives a `gen` it ignores, so a scan nominally still passes `test.Blank` — but as an explicit context-consumer, that is a sanctioned role, not the §1.1 lie. It fully disappears only if the harness ever allows target-less context-only probes.

**Upload UX/security crux (Axis 2) — does a raw client credential ever persist in Guard?** Detecting a hardcoded secret means handling the secret, so the design goal is to minimize what Guard holds:
1. **Upload-and-store** — simplest (fits Guard's artifact+job model), but a live-credential blob then sits in Guard storage → needs encryption-at-rest, short TTL, tight ACL, masked display, a "rotate these" nudge, and it is the biggest trust ask (client hands production keys to a scanner).
2. **Upload-ephemeral (recommended)** — accept the upload, scan in the worker's memory, persist **only redacted findings** (file + key name + masked value + verdict); never write raw bytes durably. The finding, not the secret, is the artifact.
3. **Local-scan, upload-findings** — run the scan where the config already lives (thin CLI / local Augustus); ship only redacted observations up. Lowest blast radius; needs a local execution path + a findings-ingest endpoint.

Secondary UX, all downstream of that choice: multi-file / zip upload (configs are scattered — host config + per-project `.mcp.json` + `.env`); a "where do I find this?" helper with per-OS Claude Desktop / Cursor / VS Code paths; masked-by-default finding display.

**Open decisions for the author:** (i) Axis 1 — (d) recommended, (c) if you'd rather not add a recon module, (b) only if a non-scored measurement is acceptable; (ii) Axis 2 — persistence model 1/2/3, recommend 2; (iii) the Augustus-side change is small (`content` already exists) — the real work is Guard-side upload + storage + ingest, so this likely splits into an Augustus recon/generator PR and a Guard plumbing PR.

# 2. RugPull tests the wrong party

A rug pull is an attack **performed by** the server, not a vulnerability **in** it: actor = the MCP server; victim = the client/host that approves once and never re-checks; the exploited weakness is the *client's* trust model (MCP04's mitigation is client-side). Via `mcp.MCP`, Augustus **is the client**, so RugPull runs the *defensive* drift-alerting control against a server — it answers "is this server malicious?", not "is this target vulnerable?" It is a detective/monitoring control, and a real malicious server evades it trivially by serving a stable definition to the scanner while mutating only for real victims.

The implementation confirms the mis-fit: it scores all drift `1.0` VULN, but a diff can't tell a malicious mutation from a benign version bump — that is monitoring output for a human to adjudicate, not a target verdict.

The faithful offensive test — is the **client** vulnerable to a rug pull — needs Augustus to play a **malicious server** (advertise benign tools, let the client approve, mutate, watch whether it re-validates). That inverts the #210 client seam, and Augustus has no server-harness. **Verdict: DEFER-CATEGORY** — either defer until that harness exists (don't approximate it by diffing a server you connect to), or re-scope + re-label as a supply-chain **drift monitor** on the observation path with no VULN verdict. (Contrast: `ResponseLeak` is coherent — there the server legitimately is the target-under-test.)

# 3. Other

## 3.1 Naming inversion (non-blocking)

Names read backwards. **`ConfigLeak` is a detector but "Leak" reads active** — the house convention for "found an artifact" is `<pkg>.<Noun>` (`apikey.ApiKey`, `productkey.Win5x5`, `knownbadsignatures.EICAR`), so **`mcpsecrets.Credential`** (or `.Secret`/`.ExposedKey`). **`SecretScan` is a probe but "Scan" reads like analysis** (the detector's job); a collection-verb name fits better (e.g. `mcpconfig.ConfigSource`). Setup is sound; only the labels are crossed. Fix before the registry names commit.

maurodelossantos-praetorian:
Thanks — this was an excellent review; the two mapping critiques are right and we've acted on all three points (`e795aa2`).

**1. SecretScan mis-mapping → adopted option (d).** Config collection is now a first-class **`recon.MCPConfig`** module (`internal/recon/mcpconfig/`) that reads `path`/`content` (dir walk, size cap, ctx, case-insensitive ext) and emits the config as `output.Observation`s. The probe is now **`mcpconfig.CredentialExposure`**, a `ContextAwareProbe` that reads those observations via `SetContext` and scores them with the detector — keeping the scored verdict, bending no abstraction, and reusing the #210 recon + `ContextAwareProbe` seam exactly as `toolsec` does (`ConfigsFrom(store)` mirrors `InventoriesFrom`). The Guard-side upload/ephemeral-storage (your Axis 2, persistence model 2 recommended) is genuinely Guard-side and is tracked separately, not in this PR.

**2. RugPull tests the wrong party → deferred (DEFER-CATEGORY).** Agreed: as Augustus-is-client, diffing a server is a detective monitor, not a target verdict (a malicious server evades it, and drift ≠ vulnerability). Removed `toolsec.RugPull` + `toolsec.ToolDrift` + `DiffTools` from this PR and opened **[LAB-4991](https://linear.app/praetorianlabs/issue/LAB-4991)** for the faithful client-side test (needs a malicious-server harness), which also captures your re-scope-as-drift-monitor alternative. (`canonicalJSON`, shared with ResponseLeak's dedup, was preserved.)

**3. Naming inversion → fixed.** Detector `mcpsecrets.ConfigLeak` → **`mcpsecrets.Credential`** (noun, per `apikey.ApiKey`); probe `mcpconfig.SecretScan` → **`mcpconfig.CredentialExposure`**. Done before the registry names shipped.

`ResponseLeak` is unchanged (per your note that it's coherent) beyond the detector-name reference — ready whenever you pick up the follow-up review.

Full suite + vet + gofmt green; the new `recon.MCPConfig` file reads keep the `filepath.Clean`+`#nosec G304` operator-supplied justification.

maurodelossantos-praetorian:
### Addressed the latest review round (`0ca6606`)

- **Codex (critical) — `responseleak.go` didn't compile.** Merging `main` (#234) extracted the tool-safety policy into the shared `internal/toolpolicy` package; `ResponseLeak` still used the removed package-local `toolPolicy`/`newToolPolicy`/`filterTools`. Migrated it to `toolpolicy.Policy` / `toolpolicy.New` / `Policy.Filter` (mirroring `injection.go`) and switched the tests to the literal config keys. No behavior change.
- **Codex-connector (P1) — "recon config not wired."** This was against the pre-merge commit `e795aa2`. The `main` merge added `ResolveReconConfig` + the `recon.settings` config section, and `createRecons` uses it — so `recon.MCPConfig` **is** configured. Verified end-to-end: `scan test.Blank --recon recon.MCPConfig --probe mcpconfig.CredentialExposure --config-file <recon.settings path>` flags a leaky config **VULN** and a clean one **SAFE**. No code change needed.
- **Claude (critical) — PR-body scope mismatch.** Corrected the description/OWASP table: this PR ships the **MCP01** config + response-leak work only; RugPull/ToolDrift/MCP04 were deferred to **LAB-4991** (not in this diff).
- **titus `ghp_` at `credentialexposure_test.go:86`** — intentional fake fixture for the detector test (dismiss, as before).

Full `go test ./...` + `go vet` + `gofmt` + gosec (G304: 0) all green.

internal/detectors/mcpsecrets/mcpsecrets.go
```diff
@@ -0,0 +1,295 @@
...
+	"strings"
+
+	"github.com/praetorian-inc/augustus/internal/detectors/apikey"
+	"github.com/praetorian-inc/augustus/pkg/attempt"
+	"github.com/praetorian-inc/augustus/pkg/detectors"
+	"github.com/praetorian-inc/augustus/pkg/registry"
+)
+
+func init() {
+	detectors.Register("mcpsecrets.Credential", NewCredential)
```
maurodelossantos-praetorian, line 29: @codex @claude @gemini — please re-review; the toolpolicy compile fix and PR-scope correction are in.

maurodelossantos-praetorian:
### Addressed the latest review round (`ef76d85`)

**Detector core (the FP/FN this feature is judged on):**
- **FN fixed** [Codex]: `isPlaceholder` matched `SafeTokens` by substring, so `DB_PASSWORD=mypassword123!` was skipped (`mypassword` ∈ SafeTokens). Now an **exact** match for config values.
- **FP fixed** [Codex-connector]: secret-key matching was an unanchored substring alternation, so `tokenizer:` matched `token`. Replaced with `isSecretKey()` matching whole key **segments** (split on `[_.-]`) plus compound forms (`api_key`, `access_token`, …). `tokenizer`/`tokenized` no longer flag.

**ResponseLeak:**
- **Recon reuse** [Codex]: now embeds `reconContext` and discovers tools via `resolveTools` (preferring a prior `recon.MCP` inventory, like `Injection`/`SSRF`) instead of always doing a live `ListTools`.
- **Full-response scan** [Codex-connector]: removed the pre-scoring 1 MiB cap — it truncated before scoring, so a secret past the boundary was missed. Now scores the full text/raw (correctness > memory-bounding for a detection probe on an operator-run scan).

**recon.MCPConfig:**
- **Symlink hardening** [Codex]: the directory walk now skips symlinked entries (a symlinked `.json`/`.env` could read outside the tree and bypass the size cap).

**Docs** [Claude]:
- Fixed the headline invocation — recon modules are configured via a `--config-file` `recon.settings.recon.MCPConfig` block, not generator-only `--config`; narrowed the goal text (no `.env`-reference resolution).

Verified: full `go test ./...` + `go vet` + `gofmt` + gosec (G304: 0) green; end-to-end CLI shows `DB_PASSWORD=mypassword123!` → **VULN** and `tokenizer: cl100k_base` → **SAFE**. (titus `ghp_` test-fixture alerts remain intentional fakes for the detector tests.)

internal/detectors/mcpsecrets/mcpsecrets.go
```diff
@@ -0,0 +1,326 @@
...
+// decompose into a single secret-word segment (e.g. "api_key", "access_token").
+var secretKeyCompound = regexp.MustCompile(`(?i)(api[_-]?key|access[_-]?key|private[_-]?key|client[_-]?secret|auth[_-]?token|access[_-]?token)`)
+
+// secretKeySeparators splits a config key / env-var name into segments.
+func secretKeySeparators(r rune) bool { return r == '_' || r == '.' || r == '-' }
+
+// isSecretKey reports whether a config key / env-var name conventionally holds a
+// credential. It matches WHOLE segments (so "tokenizer" does not match "token")
+// or the whole name against secretKeyCompound.
+func isSecretKey(name string) bool {
```
maurodelossantos-praetorian, line 133: @codex @claude @gemini — one final review please; this round fixes the placeholder-FN, the tokenizer-FP (whole-segment key matching), ResponseLeak recon-reuse + full-response scan, the symlink walk, and the docs invocation.

maurodelossantos-praetorian:
### Addressed the latest round (`7f904df`) — Codex / Gemini / Codex-connector (Claude: no criticals, LGTM)

**Detector accuracy (mcpsecrets):**
- **FN — `args` secrets** [Codex]: `{"args":["--password","…"]}` / `["--api-key=…"]` were missed despite the documented `args` coverage. Added a flag-arg signal (both `--flag=value` and `"--flag","value"` forms), gated by `isSecretKey` + `isConcreteSecret`.
- **FP — endpoint URLs** [Codex-connector]: `TOKEN_ENDPOINT=https://…/token` was flagged (URL punctuation passed the entropy check). Now bare URLs *without* userinfo are rejected (real URI creds still caught by `connCreds`).
- **FN — TOML/INI keys** [Codex-connector]: `client-secret =` / `database.password =` were missed (env key charclass excluded `-`/`.`). Broadened it.
- **FN — YAML list items** [Gemini]: `- password: X` was missed. Added an optional `- ` sequence indicator.
- **FN — long alphabetic passphrases** [Gemini]: relaxed the KV gate to accept `len>=16` non-placeholder values on high-confidence secret keys (short values still require entropy, so no new FPs).

**recon.MCPConfig hardening:**
- Skip **non-regular files** [Codex] (a FIFO/device named `config.env` would block the read).
- Resolve a **symlinked `path` root** via `EvalSymlinks` before walking [Gemini] (else zero files → false clean); entry-level symlink skip unchanged.
- **Cap collected files** at 10000 (`SkipAll` + `slog.Warn`) [Gemini].

**Cleanup:** removed the now-unused `apikey.IsSafeToken` (my earlier addition; verified unused repo-wide) [Gemini].

Verified: full `go test ./...` + `go vet` + `gofmt` + gosec (G304: 0) green; end-to-end CLI shows `--password …` args → **VULN**, `TOKEN_ENDPOINT=https://…` → **SAFE**, `client-secret =` → **VULN**. (titus `ghp_` test-fixture alerts remain intentional fakes.)

internal/detectors/mcpsecrets/mcpsecrets.go
```diff
@@ -0,0 +1,391 @@
...
```
maurodelossantos-praetorian, line 133: @codex @claude @gemini — please re-review; this round closes the args/TOML/YAML-list/passphrase FN gaps, the endpoint-URL FP, and the FIFO/symlink-root recon hardening.

maurodelossantos-praetorian:
### Addressed the latest round (`feee83c`) — Codex/Gemini/Codex-connector (Claude & Codex: no criticals)

**Detector false positives:**
- **Inline comments** [Gemini + Codex-connector]: `PASSWORD_REQUIRED=false # local dev only` was flagged — the value capture ate the `# comment`, bypassing the deny-list and passing the entropy gate. Now a whitespace-introduced `#…` comment is stripped before gating (an in-value `#` like `p#ss` is preserved).
- **Pointer/path keys** [Codex-connector]: `API_KEY_FILE=/var/run/secrets/x.json` / `client_secret_path=…` were flagged as the secret itself. `isSecretKey` now rejects keys whose last segment is a pointer word (`file`/`path`/`ref`/`location`/`dir`/`uri`/`url`/`endpoint`).
- **FN — spaces in `--flag=value`** [Codex-connector]: `--passphrase=correct horse …` only scored the first word; added a quoted-arg capture that keeps spaces.

**recon.MCPConfig:**
- **Cumulative memory cap** (50 MiB, `SkipAll`+`slog.Warn`) on top of the per-file (5 MiB) and file-count (10k) caps [Codex + Gemini].
- **Skip `node_modules`/hidden dirs** in the walk (never the root) [Gemini].

**Deliberately not changed:** re-capping `ResponseLeak`'s stored Text/Raw — that truncation was removed last round to avoid false negatives; a single tool response is server-bounded (the tens-of-GiB risk is the dir walk, now capped).

Verified: full `go test ./...` + vet + gofmt + gosec (G304: 0) green; end-to-end shows the comment and pointer-key cases **SAFE** and a real `ghp_` token **VULN**.

internal/detectors/mcpsecrets/mcpsecrets.go
```diff
@@ -0,0 +1,442 @@
...
+	"dir": true, "uri": true, "url": true, "endpoint": true,
+}
+
+// secretKeySeparators splits a config key / env-var name into segments.
+func secretKeySeparators(r rune) bool { return r == '_' || r == '.' || r == '-' }
+
+// isSecretKey reports whether a config key / env-var name conventionally holds a
+// credential. It matches WHOLE segments (so "tokenizer" does not match "token")
+// or the whole name against secretKeyCompound.
+func isSecretKey(name string) bool {
```
maurodelossantos-praetorian, line 133: @codex @claude @gemini — please re-review; this round fixes inline-comment & pointer/path-key false positives, a --flag=value spaces FN, and adds a cumulative recon memory cap + node_modules/hidden-dir skip.

maurodelossantos-praetorian:
### Addressed the latest round (`bea8b68`) — Gemini / Codex-connector (Claude & Codex: no criticals)

**Detector accuracy:**
- **camelCase keys** [Gemini + Codex-connector]: `dbPassword`/`apiToken` collapsed to a single segment (FN); `isSecretKey` now splits on camelCase transitions too, so those match — and `apiKeyFile` correctly resolves to a pointer key (fixing that FP).
- **Escaped quotes in arg regexes** [Gemini]: `["--password","S3c\"ret"]` was truncated; now uses the escape-aware capture like `jsonKV`.
- **Comment-strip vs quoted values** [Gemini]: `DB_PASSWORD="my #secret"` was truncated to `my`; the stripper is now quote-aware (strips comments only from unquoted values).

**recon.MCPConfig:**
- **Hidden-dir over-skip** [Codex-connector]: last round's `.`-prefix skip dropped `.config`/`.cursor`/`.vscode` — exactly where MCP configs live. Now skips only `node_modules` and `.git`.

**ResponseLeak:**
- **Deterministic attempt order** [Gemini]: params are sorted before deriving arg cases (was Go map-iteration order).
- **Generous response cap** [Gemini + Codex]: re-introduced a **10 MiB** cap (vs the 1 MiB that caused a false negative and was removed) — bounds memory/report size while staying far above where real leaked credentials appear. Documented the sizing so it stops flip-flopping.

Verified: full `go test ./...` + vet + gofmt + gosec (G304: 0) green; end-to-end shows camelCase `dbPassword` in `.cursor/mcp.json` → **VULN** (both the camelCase and hidden-config-dir fixes) and `apiKeyFile` pointer → **SAFE**. (titus `ghp_` test-fixture alerts remain intentional fakes.)

internal/detectors/mcpsecrets/mcpsecrets.go
```diff
@@ -0,0 +1,501 @@
...
+			start = i
+		}
+	}
+	return append(out, strings.ToLower(string(runes[start:])))
+}
+
+// isSecretKey reports whether a config key / env-var name conventionally holds a
+// credential. It matches WHOLE segments (so "tokenizer" does not match "token")
+// or the whole name against secretKeyCompound.
+func isSecretKey(name string) bool {
```
maurodelossantos-praetorian, line 133: @codex @claude @gemini — please re-review; this round fixes camelCase key matching, hidden-config-dir collection (.cursor/.config), arg-value escaping, quote-aware comment stripping, deterministic attempt order, and a 10 MiB response cap.

maurodelossantos-praetorian:
### Addressed the latest round (`dafd2cb`) — Codex & Claude: no critical issues (Claude: LGTM pending human review)

One actionable finding this round:

- **Raw-payload cap applied as bytes** [Codex-connector]: `callOne` converted `res.Raw` to a string *before* `truncateResponse` capped it, so a huge/hostile raw payload still forced one unbounded allocation — defeating the 10 MiB cap. Added `truncateResponseBytes` to cap the `[]byte` **before** the string conversion, and dedupe against the equally-capped Text form. New test asserts a secret within the cap is still scored and no stored output exceeds cap+marker.

Full `go test ./...` + `go vet` + `gofmt` + gosec (G304: 0) green.

Both named reviewers reported **no critical issues** this round (and the prior one); remaining items across recent rounds have been narrow config-syntax/hardening edge cases, all addressed. From my side this reads as **converged and ready for human review**.

internal/probes/toolsec/responseleak.go
```diff
@@ -0,0 +1,259 @@
...
+		return s
+	}
+	return s[:maxResponseBytes] + "…[truncated]"
+}
+
+// truncateResponseBytes caps a raw byte payload to maxResponseBytes and only then
+// converts it to a string. Truncating before the []byte→string conversion avoids
+// materializing the entire (possibly huge/hostile) raw payload as a string first,
+// so the cap actually bounds the allocation rather than merely the stored output.
+func truncateResponseBytes(b []byte) string {
```
maurodelossantos-praetorian, line 42: @codex @claude @gemini — please re-review; this round caps the raw tool-response payload as bytes before the string conversion so the 10 MiB bound prevents an unbounded allocation.

maurodelossantos-praetorian:
### Addressed the latest round (`4676f70`)

Two small, clean true positives applied:

- **Username-less connection strings** [Gemini]: `connCreds` required ≥1 char before the `:`, missing `redis://:S3cr3tP@ss@host`. Allowed an empty username (`+`→`*`). Regression-checked: `scheme://host:443@…` still **SAFE** (the all-digit password guard holds), existing `postgres://admin:pass@…` still VULN.
- **Config-read size race (TOCTOU)** [Codex-connector]: both read sites now use `readCapped` (`os.Open` + `io.LimitReader` to `maxFileSize+1`), so a file that grows/swaps between the `Size()` check and the read can never buffer past the cap.

**Deliberately not changed this round — flagged for human review, not bugs:**
- Codex "critical" (*truncation drops secrets past 10 MiB*): the detector scores `Attempt.Outputs`, so "store bounded" and "score full" can't both hold without an architectural score/store split — out of scope here. 10 MiB is a defensible cap (real leaked credentials appear in the first KB). This has oscillated across rounds (1 MiB → removed → 10 MiB → byte-level); holding it stable.
- Bounding the response *before the generator materializes it* belongs to the `mcp.MCP` generator / `ToolInvoker` layer, not this probe.
- empty-recon-store → SAFE informational: Claude flagged it as a deliberate, documented UX choice for a human to confirm.

Verified: full `go test ./...` + vet + gofmt + gosec (G304: 0) green; end-to-end shows `redis://:pass@host` → **VULN** and `scheme://host:443@` → **SAFE**.

Both named reviewers reported **no critical correctness issues** the last two rounds (Codex's latest "critical" is the cap tradeoff above, not a defect). This reads as **converged**; the open items are human judgment calls best settled in review.

internal/detectors/mcpsecrets/mcpsecrets.go
```diff
@@ -0,0 +1,505 @@
...
+var yamlKV = regexp.MustCompile(`(?m)^\s*(?:-\s+)?([A-Za-z0-9_.\-]+)\s*:[^\S\r\n]*(.+?)[^\S\r\n]*$`)
+
+// connCreds matches a credential embedded in a URI userinfo section (the
+// "user:secret@host" form). The secret capture is greedy up to the LAST '@'
+// before the host, so a secret that itself contains '@' is captured in full.
+// The username quantifier is '*' (not '+') so a userinfo credential that OMITS
+// the username (e.g. "redis://:secret@host") is still caught; the all-digit
+// password guard on the connCreds signal keeps a URL port ("host:443@rest")
+// from being mistaken for a password.
+var connCreds = regexp.MustCompile(`[a-zA-Z][a-zA-Z0-9+.\-]*://[^/\s:@]*:([^/\s]+)@`)
```
maurodelossantos-praetorian, line 189: @codex @claude @gemini — please re-review; this round catches username-less connection strings and closes a config-read size-race (TOCTOU). The response-cap tradeoff is intentionally held stable (see the PR comment).

praetorian-farida (approved):
**LGTM — approving.** The revision addresses the review cleanly:

- Config scanning re-homed to `recon.MCPConfig` (collect) + `mcpconfig.CredentialExposure` context-aware probe (score) — keeps the verdict, no faked target, and the `mcp.config` observation round-trips correctly through the store (producer/reader co-located, single source of truth).
- `RugPull` deferred, as discussed (the offensive test needs a malicious-server harness we don't have yet).
- Detector renamed to `mcpsecrets.Credential`, plus the FP/FN hardening pass.

One ask before you merge: **please run it against a live Guard instance first.** The new recon→context-probe flow and the config-input path (upload/`content`) are the parts that haven't been exercised end-to-end through Guard, and that's where the plumbing is most likely to need small adjustments (how the uploaded config reaches the recon module's input, and how findings surface). Approving so this isn't blocked — just want the Guard run to confirm before it lands, in case changes are needed.

**Once working in Guard, please merge.**

maurodelossantos-praetorian:
@praetorian-farida — tested this end-to-end against a local Guard deploy (Floci), with Guard's Augustus pin bumped to this branch. Both probes check out, vulnerable vs. safe:

- **`toolsec.ResponseLeak`** — leaky MCP server → risk emitted; clean server → nothing.
- **`mcpconfig.CredentialExposure`** — leaky config (fed through `recon.MCPConfig` content) → risk; clean config → nothing.

The recon→context-probe flow and the config-input/upload path both work through Guard, and findings surface with CVSS/CWE/framework tags. That covers the "run it in Guard first" ask — good to merge from my side.

Next up is the Guard-side UI work (config-input UX, simplifying the Augustus main tab/run, and curated finding definitions), which I'll carry separately.

PR #261, opened 2026-07-21
feat(mcptool): blind & in-band OS command-injection probing for MCP tools (LAB-4929)

maurodelossantos-praetorian:
## What & why

Extends `mcptool.Injection` to detect **OS command injection** (OWASP **MCP05**, ~43% ecosystem prevalence — CVE-2025-6514, CVE-2025-53355) in directly-invokable MCP tools. Closes the biggest gap in the existing probe: it only caught injection that **reflects a computed result in-band** (eval/SSTI arithmetic canary). Real command-injection sinks are frequently **blind** — the command runs but nothing comes back to the client.

Implements LAB-4929 (MCP command injection & execution). Pure extension of the merged probe/detector — no new infrastructure; reuses the OOB collector from `mcptool.SSRF`.

## How it works

- **New payload family (`oobCmdFormats`)**: shell-metacharacter payloads embedding a unique canary URL — statement separators (`;`, `|`, `&`, `&&`), command substitution (`$(…)`, backticks), `wget` fallback, newline injection, and single/double-quote breakout.
- The probe stands up the built-in OOB collector, injects **both** families (in-band arithmetic + out-of-band shell) into every string parameter, then records a per-attempt **OOB-callback** flag after a grace period.
- The **detector** flags an attempt on **either** signal:
  - a reflected canary marker → in-band / non-blind, or
  - an OOB collector callback → **blind** command injection (fires even when the tool returns nothing).
- New metadata keys: `injection_oob_callback`, `injection_oob_url`.

## Design guarantees

- **No false positives from echo**: the marker is a computed value / collector body that never appears in the payload text, so a tool that merely reflects input can't trip either signal.
- **Command-injection vs SSRF**: the URL sits behind shell metacharacters; a plain URL-fetch can't parse it, so a callback means a shell actually ran.
- Respects the existing destructive-tool safety gate and shared-recon reuse.

## Validation (real targets, over SSE)

| Target | Sink | Expected | Result |
|---|---|---|---|
| Controlled stub `net_diagnostic_quiet` | real shell, **output discarded** | VULN (blind) | ✅ 7 caught via **OOB callback only** |
| Controlled stub `net_diagnostic` | real shell, output returned | VULN (reflected) | ✅ 8 caught |
| DVMCP challenge 8 | real Python `eval()` | VULN (in-band) | ✅ bare-eval canary fires |
| DVMCP challenge 9 | **simulated** exec (echoes command) | SAFE (FP control) | ✅ 0/108 |
| Controlled stub `safe_echo` | reflection only | SAFE (FP control) | ✅ 0 |

> Note: DVMCP's command-injection challenges either use a Python `eval()` sink (challenge 8 — exercises the in-band path) or only *simulate* execution by echoing the command string (challenge 9). Neither implements a real **shell** sink, so the blind-OOB path is validated against a controlled real-shell stub (same legacy HTTP+SSE transport DVMCP uses). Challenge 9 doubles as a false-positive control.

## Tests

Probe + detector unit tests for the blind, reflected, and echo-false-positive paths. `go test -race`, `go vet`, and `golangci-lint` (pinned v2.12.2) all clean.


maurodelossantos-praetorian:
Thanks for the reviews — worked through all five. Pushed `f9d4942`. Summary of what I changed, and what I deliberately didn't.

## Fixed (true positives)

**1. False positive from URL-fetching / SSRF-like tools (Codex, critical).** Correct and important — and it was actually worse than described: the signal could be forged via *both* the callback *and* the reflected collector body (the collector answers every `/oob/<token>` hit with its marker). A tool that extracts a URL from a text argument and fetches it would trip either path without any shell running.

Fix, in two parts:
- **`shellProofURL`**: splice an empty command substitution `$()` into the middle of the token. A POSIX shell evaluating the argument collapses `$()` to nothing and requests the real `/oob/<token>` path — *command-execution-specific proof*. A tool that fetches the literal URL from the text instead requests a `…$()…` path whose token doesn't match the tracked one, so an SSRF/link-fetch sink can no longer masquerade as command injection.
- **Dropped the collector-marker reflection signal for OOB attempts.** The callback alone catches both blind *and* non-blind command injection (a shell that runs `curl` fires the callback whether or not it returns the body), so removing reflection loses no true positives and closes the second FP vector.
- Added the negative test Codex asked for: a URL-fetching tool now yields **0** injection findings. Verified on a real target too — adding an SSRF-like `read_webpage` tool to the validation server left the injection count unchanged at 15.

**2. Finding masked by error status (Gemini, critical).** Confirmed against `pkg/results/results.go:133` — `Verdict()` returns `"error"` for `StatusError` regardless of detector score. A blind payload that executes but then hangs/errors the tool call would be silenced. Fix: after the callback wait, a callback-firing attempt still in `StatusError` is reverted to complete (original error preserved in `toolsec.injection_oob_call_error`) so it surfaces as VULN.

**3. Context cancellation (Gemini, minor).** The `Probe` loop now stops issuing new calls once `ctx` is done, while still reconciling callbacks for what was already sent.

## Not changed (with rationale)

**Intra-probe `errgroup` fan-out (CodeRabbit).** Deferring: the sibling `toolsec.SSRF` probe is deliberately sequential with the same structure, and the scanner already enforces bounded concurrency (10 goroutines) at the probe/attempt level. Parallelizing one probe but not the other adds inconsistency and OOB-token/rate-limit bookkeeping for a scan-time win that's better taken as a single consistent refactor of both `toolsec` probes.

**PowerShell payloads (Gemini).** Deferring: `powershell` quoting doesn't share the `$()` no-op semantics the FP fix relies on, so adding it now would either reintroduce the ambiguity or need a separate proof mechanism. `curl` ships on Windows 10+, and on `cmd.exe` the `$()` token simply doesn't collapse — a *false negative* (safe direction), not a false positive. Broader Windows-shell coverage is worth a dedicated follow-up rather than bolting it onto this fix.

## Re-validation (real MCP servers, over SSE)

`subprocess(shell=True)` sink → 8 exec + **7 blind callback-only** catches (unchanged); URL-fetch tool → **0** injection findings (FP fixed); DVMCP #8 (real `eval()`) → 1 VULN; DVMCP #9 (simulated) → 0. `go test -race`, `go vet`, `golangci-lint` v2.12.2 all clean.

internal/probes/mcptool/injection.go
```diff
@@ -119,3 +203,67 @@ func (p *Injection) callOne(ctx context.Context, inv types.ToolInvoker, tool, pa
...
+// execution reproduces the tracked token. It splices an empty command
+// substitution ("$()") into the middle of the token: a POSIX shell evaluating the
+// argument collapses "$()" to nothing, requesting the real /oob/<token> path the
+// collector tracks — command-execution-specific proof a plain URL fetch cannot
+// forge. A tool that instead extracts and fetches the literal URL from the
+// argument text requests a "...$()..." path, whose token does not match the
+// tracked one, so an SSRF / link-fetch sink cannot masquerade as command
+// injection. (On Windows cmd.exe "$()" is not a no-op, so cmd-only sinks may be
+// missed — a false negative, the safe direction.)
+func shellProofURL(url, token string) string {
```
maurodelossantos-praetorian, line 254: Re-review request for the pushed fixes: this `shellProofURL` helper is the core of the false-positive fix — it makes the OOB callback command-execution-specific (only a real shell collapses the `$()` back to the tracked token). Along with dropping the reflection signal, the error-status recovery, and the ctx checks, could you take another pass? @codex review @gemini review @claude please re-review

maurodelossantos-praetorian:
Second round addressed in `d1aa660`. All three re-reviews came back with no critical issues (Codex: LGTM; Claude & Gemini: no critical), so these are the non-blocking follow-ups:

- **Detector doc accuracy (Claude + Gemini):** `InjectionDetector.Detect` no longer claims the reflected-marker path covers "returned the collector's body (non-blind command injection)." Since `callOOB` intentionally carries no reflection marker (the callback covers blind *and* non-blind, and reflecting the body would false-positive on SSRF-like tools), the comment now matches the code.
- **Context cancellation (Gemini):** moved the `ctx.Err()` check inside the payload loops (was per-param), so cancellation stops immediately rather than emitting a burst of erroring attempts for the rest of a param's payloads.
- **Test clarity (CodeRabbit):** `TestInjection_DetectsCommandInjectionViaOOBCallback` now uses `t.Run` subtests.

No behavior change to detection — re-verified 15 VULN (8 exec + 7 blind callback-only) on the real-shell stub. `go test`, `go vet`, `golangci-lint` v2.12.2 all clean.

maurodelossantos-praetorian:
Rebased onto latest `main` by merging `origin/main` (`5b6c8d0`). `main` renamed the `toolsec` package to **`mcptool`**, so this probe is now **`mcptool.Injection`** (metadata keys `mcptool.injection_oob_callback` / `mcptool.injection_oob_url`). Also regenerated the new `pkg/register` (`make generate`; `make generate-check` passes). Conflicts resolved; PR is MERGEABLE. Re-validated over SSE post-merge: real-shell sink **15 VULN** (8 exec + 7 blind callback-only), DVMCP #9 simulated **0**. build / go test -race / lint / generate-check all green.

internal/probes/mcptool/injection.go
```diff
@@ -119,3 +208,67 @@ func (p *Injection) callOne(ctx context.Context, inv types.ToolInvoker, tool, pa
...
+// Detection is by out-of-band callback ONLY — deliberately not by reflecting the
+// collector body. The collector answers every /oob/<token> hit with the same
+// marker, so a tool that merely extracts and fetches the URL from the argument
+// text (an SSRF / link-preview sink, not a command sink) would get that marker
+// back and be misclassified as command injection. The callback path avoids this:
+// the URL's token carries a shell-proof marker (see shellProofURL) that only a
+// real shell resolves to the tracked token, and the callback catches BOTH the
+// blind case (nothing returned) and the non-blind case (a shell ran curl and the
+// callback still fired), so dropping reflection loses no true positives.
+func (p *Injection) callOOB(ctx context.Context, inv types.ToolInvoker, tool, param string, params []paramInfo, format, canaryURL string) *attempt.Attempt {
```
maurodelossantos-praetorian, line 224: Post-rebase review request (`5b6c8d0`, now on latest main; package renamed toolsec→mcptool, so this is `mcptool.Injection`). This `callOOB` is the blind command-injection path: it fetches a shell-proof canary URL and detection is by OOB callback only. Please re-review the whole probe/detector after the merge and flag any CRITICAL or MAJOR issues (evasions, false positives/negatives, correctness) before merge. @codex review @gemini review @claude please re-review @coderabbitai review

maurodelossantos-praetorian:
## Ready for human merge review ✅

Rebased onto latest `main` (package `toolsec`→`mcptool`; this probe is now **`mcptool.Injection`**), conflicts resolved, and the post-rebase review round addressed. All checks green: `build`, `go test -race`, `golangci-lint`, `make generate-check`.

**What this PR does:** adds server-integrated OS **command-injection** testing (OWASP MCP05) to `mcptool.Injection` — shell-metacharacter payloads that catch **blind** command injection (no output returned) via an out-of-band callback, alongside the existing in-band arithmetic canary. Detection is proof-based (computed marker or a private per-token OOB callback), so it structurally resists false positives.

**Addressed in the latest commit** (the only two review items, both minor):
- **Codex P1:** quote-breakout payloads now end with ` #` so they terminate the wrapper's trailing quote and actually fire against sinks that quote user input.
- **Gemini:** OOB fetches carry a short timeout (`curl -m 3` / `wget -T 3`) so a blackholed collector can't stall the probe.

**Validation:** re-verified over SSE against a real `shell=True` sink — command injection caught (15 VULN: 8 exec + 7 blind callback-only), **0 false positives** on benign URL-fetch/echo tools, and DVMCP #9 (simulated exec) stays clean. Earlier review rounds (Claude, Codex summary) were LGTM with no critical issues.

praetorian-farida (approved):
Nice work — this is a well-constructed extension. The blind-injection path is the right gap to close (the in-band arithmetic canary genuinely can't see a discarded-output sink), and the engineering around it is careful: the two payload families are cleanly separated, the destructive-tool gate is applied before anything is invoked (`injection.go:108`), and the test matrix covers the paths that matter — blind, non-blind, and both false-positive controls.

Two **non-blocking** notes, neither gating this approval:

1. **Fetch timeout isn't coupled to the grace period.** `oob_wait_seconds` is configurable (default 3), but the injected `curl -m 3` / `wget -T 3` timeouts are hardcoded at 3s (`oobCmdFormats`, `mcptool.go`). If an operator raises `oob_wait_seconds` for a slow target, the injected fetch still gives up at 3s, so the longer grace period doesn't actually buy a slow sink more time to call back. Consider deriving the fetch timeout from the wait, or giving it its own knob.

2. **PR-body naming.** The description refers to `toolsec.Injection`; the registered name is `mcptool.Injection`. Worth fixing the body so it matches the shipped name.

One design gap I looked at and deliberately did **not** hold against this PR: the OOB collector defaults to a loopback bind, so against a *remote* target with `oob_base_url` unset, a real blind sink produces no callback and scores safe — a false negative driven by misconfiguration rather than a clean target. It's inherited from the shared collector (the SSRF probe has the identical property), the config knobs to fix it already exist (`oob_base_url` / `oob_listen`), and the in-band path is unaffected. I've filed **LAB-5092** for the proper fix — a modular `ProbePreflight` advisory (target-aware warn when remote + unset, using the existing `types.MCPEndpoint`), reusable across `Injection` and `SSRF` — so it doesn't need to ride in this PR.

Approving. 🚢

PR #262, opened 2026-07-21
feat(toolsec): static server-integrated tool-poisoning probe (LAB-4460)

maurodelossantos-praetorian:
## What & why

Adds **`toolsec.ToolPoisoning`** + the **`toolsec.ToolPoisoning` detector** — server-integrated **tool-poisoning** detection for MCP targets. Tool poisoning is the **#1 MCP attack class** (OWASP **MCP03** + **MCP10**; 36–93% ASR across 20 LLMs; the vector behind the GitHub MCP incident, Cursor CVE-2025-54136, and the postmark-mcp supply-chain attack). This closes the biggest remaining gap in Augustus's server-integrated MCP coverage.

Implements the **static** half of LAB-4460.

## How it works

The probe reads the **target server's own advertised metadata** — tool descriptions, input schemas, server instructions, and resource/prompt descriptions — from a shared **`recon.MCP` inventory** (scan once, reuse everywhere) or a live **`ListTools`** fallback, and emits one attempt per scanned item. The detector deterministically flags hidden instructions aimed at the host LLM:

- **instruction-smuggling tags** — `<IMPORTANT>` / `<HIDDEN>` / `<system>` / …
- **model-directed directives** — secrecy (`do not mention`), deception (`present it as if`), cross-resource exfiltration (`access the resource at … include it in your response`), tool-shadowing (`always use X instead`), and the two-step `before … you must first …` hijack. Deliberately high-precision: honest parameter guidance like *"you must provide a valid city"* does **not** trip it.
- **invisible / bidi Unicode smuggling** — zero-width and bidi-override characters (explicit code points, `check_hidden_unicode` toggle).

**Server-integrated, not model-facing.** The poison comes from the *real target's* catalog — which distinguishes this from the existing synthetic model-facing probes (`toolcoercion.MCPToolPoison`, `tool.SchemaMutation`) that bake poisoned defs into a prompt. No LLM required; deterministic and offline once recon has run. A target with neither recon nor a tool surface **fails loud**, never a silent SAFE.

## Validation (real MCP servers, over SSE)

DVMCP ships its `<IMPORTANT>`/`<HIDDEN>` poisoning **only in the stdio `server.py` variants**; every `server_sse.py` variant — the only ones Augustus's remote-only generator can reach — has it stripped (challenge #2's SSE variant is a command-injection tool, not a poisoned one). So I validated against a controlled SSE target that **reproduces DVMCP #2's canonical poisoned descriptions verbatim**, plus a benign FP control:

| Target | Expect | Result |
|---|---|---|
| Poisoned target (`get_company_data` `<IMPORTANT>`, `search_company_database` `<HIDDEN>`, benign `get_weather`) | 2 VULN, rest SAFE | ✅ 2 VULN (the poisoned descriptions), `get_weather` + all input-schemas SAFE |
| DVMCP #9 (benign descriptions) | 0 findings | ✅ 0/4 |

## Tests

Probe (recon-path attempt generation, live fallback, fail-loud) + detector (8 true-positive signal classes, 5 false-positive guards, Unicode toggle) unit tests. `go test -race`, `go vet`, `golangci-lint` v2.12.2 all clean.

## Scope / follow-up

This PR is the **static** analysis half. The **dynamic follow-through** — present the real poisoned tool defs to a host LLM and judge (via `judge.Judge`) whether it obeys the hidden instructions — is deliberately deferred: it needs a host-LLM generator (Bifrost/API key) that can't be validated in this local environment, and shipping unvalidated detection would be worse than scoping it as a second slice. The static probe is the higher-signal, always-available, deterministic capability and stands on its own.


maurodelossantos-praetorian:
Addressed the review in `792a6c6`. All three re-reviews were no-critical (Claude explicit LGTM); fixes:

- **Nil-map panic (Codex, critical):** the detector records its signal via `a.WithMetadata` (nil-safe) instead of a raw `a.Metadata[...]` write, so `Detect` can't panic on a caller-constructed attempt with no Metadata map. In practice the probe always uses `attempt.New` (which initializes it), but the detector should be robust to any attempt.
- **`Title` false-negatives (Codex):** now scans tool / resource / resource-template / prompt `Title` fields **and** tool annotation titles — poison hidden in a title was previously missed.
- **Emoji-ZWJ / script-joining false positives (Gemini):** dropped ZWNJ (U+200C) and ZWJ (U+200D) from the invisible-Unicode set — they're required in emoji ZWJ sequences (👩‍💻) and Arabic/Persian/Indic joining, so a benign internationalized/emoji description would have false-fired. Kept ZWSP / WORD JOINER / BOM / bidi controls, which have no benign use in tool metadata. Regression test added.
- **Tightened widest-recall patterns (Claude, optional):** `pretend` now matches only directed forms; the bare `as an ai` matcher removed. Added FP guards for both.
- **`jsonString` marshal-error (Gemini, minor):** returns `""` rather than Go's non-JSON map format.

**Not changed:** Claude's empty-catalog note (zero attempts, no error) — an empty catalog genuinely has nothing to scan, so erroring would be wrong; the fail-loud path is correctly reserved for "no recon AND not tool-invokable."

Re-validated over SSE: poisoned target still **2 VULN**, DVMCP #9 still **0**. `go test -race`, `go vet`, `golangci-lint` v2.12.2 all clean.

internal/detectors/mcptool/toolpoisoning.go
```diff
@@ -0,0 +1,133 @@
...
+			if _, ok := a.GetMetadata("toolsec.poison_signal"); !ok {
+				a.WithMetadata("toolsec.poison_signal", sig)
+			}
+		}
+	}
+	return scores, nil
+}
+
+// classify returns a short signal label if the text is poisoned, else "".
+func (d *ToolPoisoningDetector) classify(text string) string {
```
maurodelossantos-praetorian, line 118: Re-review request for the pushed fixes in `792a6c6` (nil-map guard, Title-field scanning, emoji-ZWJ false-positive fix, tightened patterns). This `classify` function is the core detection logic — could you take another pass to confirm nothing critical remains? @codex review @gemini review @claude please re-review

maurodelossantos-praetorian:
Second review round addressed in `5b5c42f`. No critical issues this pass; these harden precision and align the two code paths:

**False-positive fixes (Gemini critical + Claude):**
- Removed the standalone `include … in your response` matcher — it flagged honest formatting guidance ("Search the DB and include the results in your answer"). The `access <sensitive resource> at …` exfiltration regex already carries the real signal, and DVMCP still fires via the `<IMPORTANT>`/`<HIDDEN>` tag + secrecy directives.
- `without the user` now requires a secrecy object (`knowing` / `being told` / …), so "fails without the user token" no longer trips it.
- Dropped `elevated permissions` and `system note` — both appear in honest admin tool docs ("requires elevated permissions"). Kept the stronger `admin override` / `system override` fake-authority tells.

**Coverage / hardening:**
- Live `ListTools` fallback now scans the annotation title too (Codex) — no-recon runs now match the recon path; regression test added.
- Skip the annotation-title attempt when it equals the top-level title (Gemini, dedup).
- Cap scanned metadata at 64 KiB per item — the target server is untrusted (Gemini).
- Warn instead of a silent empty pass when a recon inventory decodes but carries no scannable metadata (Claude).

Added FP-guard tests (without-the-user-token, include-results formatting, elevated-permissions doc, system-note) plus a directed "without the user knowing" true positive. Re-validated over SSE: poisoned target still **2 VULN**, DVMCP #9 still **0**. `go test -race`, `go vet`, `golangci-lint` v2.12.2 all clean.

internal/detectors/mcptool/toolpoisoning.go
```diff
@@ -0,0 +1,142 @@
...
+// PoCs and DVMCP use to hide directives from the human UI while the LLM still
+// reads them.
+var poisonTagRE = regexp.MustCompile(`(?i)<\s*/?\s*(important|hidden|secret|system|admin|instructions?|ai[_ ]?instructions?|assistant|inject)\s*>`)
+
+// poisonDirectiveREs match model-directed instructions that belong in a prompt,
+// not in a tool/resource description. They are deliberately high-precision —
+// secrecy, deception, cross-resource exfiltration, tool-shadowing, and the
+// "before doing X you must first do Y" two-step hijack — so honest parameter
+// guidance ("you must provide a valid city") does NOT trip them.
+var poisonDirectiveREs = []*regexp.Regexp{
```
maurodelossantos-praetorian, line 28: Re-review request for `5b5c42f`: this round tightened the poisoning directive patterns to kill the false positives from the last pass (dropped standalone `include … in your response`, scoped `without the user` to a secrecy object, removed `elevated permissions`/`system note`), aligned the live-fallback annotation-title scan, and capped scanned metadata size. Could you take another pass and confirm no critical issues remain — especially any residual false positives on benign tool descriptions? @codex review @gemini review @claude please re-review

maurodelossantos-praetorian:
Third round addressed in `cc7ba3b`. Codex flagged one real false negative (critical); Claude/Gemini no-critical.

- **Schema tag-poisoning false negative (Codex, critical):** input schemas are now **decoded** and scanned as their extracted string content (keys, descriptions, enum values) rather than as a raw JSON string. JSON-hex-escaped tags — `<IMPORTANT>`, as Go's `json.Marshal` and some servers emit — are unescaped to a literal `<IMPORTANT>` before matching, so tag-based schema poisoning is no longer hidden by the wire encoding. Bonus: drops the JSON punctuation the heuristics don't need. Regression test builds the schema via `json.Marshal` (which escapes the brackets) and asserts the decoded scan recovers the literal tag.
- **Tool `title` missing on live fallback (Codex + Gemini + Claude):** the no-recon `ListTools` path now scans a top-level tool `title` too, matching recon-path coverage. (Resources/prompts stay recon-only — `ListTools` doesn't expose them; recon is the recommended path.)
- **`pretend` regex boundary (Gemini):** the trailing `\b` now bounds the whole alternation, so "pretend nothingness" can't match.

Re-validated over SSE: poisoned target still **2 VULN**, DVMCP #9 still **0**. `go test -race`, `go vet`, `golangci-lint` v2.12.2 all clean.

internal/probes/mcptool/toolpoisoning.go
```diff
@@ -0,0 +1,276 @@
...
+
+// schemaText extracts the scannable human-readable content of a tool input
+// schema — every JSON key, string value, and enum entry — as newline-joined
+// plain text. It accepts either raw schema bytes (recon path) or an
+// already-decoded value (live ListTools path). Decoding first is deliberate: it
+// unescapes any JSON-hex-encoded characters (`<IMPORTANT>` -> literal
+// `<IMPORTANT>`) so tag-based schema poisoning is caught rather than hidden by
+// the wire encoding, and it drops JSON punctuation the poison heuristics don't
+// need.
+func schemaText(v any) string {
```
maurodelossantos-praetorian, line 255: Re-review request for `cc7ba3b`: this round fixed the schema tag-poisoning false negative — `schemaText` (this function) now decodes the input schema and scans its extracted string content, so JSON-hex-escaped tags (`<IMPORTANT>`) are normalized to literal tags before matching. Also added top-level tool `title` scanning on the live-fallback path and tightened the `pretend` boundary. Could you take another pass and confirm no critical issues remain — false positives on benign metadata, false negatives on smuggled poisoning, or any correctness gaps? @codex review @gemini review @claude please re-review

maurodelossantos-praetorian:
Fourth round addressed in `62f71e5`. No critical from any reviewer this pass; these harden the two Codex security notes + Gemini polish:

- **Truncation evasion (Codex):** oversized untrusted metadata was capped to 64 KiB and the prefix scanned as if it were the whole item — a server could pad past the cap and hide poison in the tail. Truncated items are now marked **inconclusive**; the detector scores them `InconclusiveScore` (review) instead of a silent SAFE, so the evasion surfaces for a human. `collectStrings` also short-circuits once it exceeds the cap (bounds allocation on a huge schema — Gemini), and `capScan` trims to a valid UTF-8 boundary (Gemini).
- **Fallback tool-title false negative (Codex):** you were right that the mock masked it — `toolsToMaps` (the live `ListTools` projection) didn't emit `title`, so the fallback title scan only worked against a mock, not the real `mcp.MCP` generator. `toolsToMaps` now carries `title`, matching the recon inventory; added a generator-level regression test (`TestToolsToMaps_IncludesTitle`).
- **`schemaText` readability (Gemini):** unmarshal into a fresh `var parsed any`.
- **server_instructions matching directive regexes (Claude):** by design — server instructions are a genuine poisoning vector — so left as-is.

Added tests: detector inconclusive-on-truncation (benign→0.5, poison-in-prefix→1.0), probe oversized-metadata marking, generator title projection. Re-validated over SSE: poisoned target still **2 VULN**, DVMCP #9 still **0**. `go test -race`, `go vet`, `golangci-lint` v2.12.2 all clean.

internal/probes/toolsec/toolpoisoning.go
```diff
@@ -0,0 +1,304 @@
...
+		a.Metadata[attempt.MetadataKeyInconclusiveReason] = fmt.Sprintf("%s exceeded the %d-byte scan cap; only the prefix was scanned", location, maxScanLen)
+	}
+	a.AddOutput(scanned)
+	a.Complete()
+	return a
+}
+
+// capScan bounds an untrusted metadata string to maxScanLen, trimming back to a
+// valid UTF-8 boundary, and reports whether it had to truncate.
+func capScan(s string) (string, bool) {
```
maurodelossantos-praetorian: Re-review request for `62f71e5`: this round closed the two Codex security notes — oversized untrusted metadata is now capped (`capScan`, this function, rune-safe) and marked inconclusive so a padding-evasion surfaces for review instead of a silent SAFE, and the live-fallback `title` false negative is fixed at the source (`toolsToMaps` now emits `title`). Could you take another pass and confirm nothing critical remains — evasions, false positives, false negatives, or correctness gaps? @codex review @gemini review @claude please re-review

maurodelossantos-praetorian:
Fifth round addressed in `4191d4f`. This round found that the cap+inconclusive machinery I added last round was **net-negative** — you were both right, and I've removed it in favor of full scanning:

- **`capScan` O(N²) + full bypass (Gemini, critical):** `utf8.ValidString` re-checks the whole prefix each iteration, and a single leading invalid UTF-8 byte (`\xFF`) made the loop strip the string to `""` — hiding the payload entirely. A self-inflicted evasion. **Removed.**
- **0.5 inconclusive renders SAFE (Codex, critical):** confirmed at `results.go:145,151` — only `> 0.5` is vuln and the `== 0.5` review band is gated on `isVisibleMultimodal`, which tool-poisoning attempts aren't. So truncated items passed as SAFE and the padding evasion persisted. **Removed the inconclusive path.**

**Fix:** `mkAttempt` now scans the **full** text. Plain metadata is already resident in memory from recon, and the detector's RE2 patterns are linear-time (no catastrophic backtracking), so full scanning is O(n) with **no evasion** — poison anywhere in a description/title/instruction is caught. Only schema *extraction* stays size-bounded, now via `writeBounded` which strictly caps each individual write (addresses Codex's "allocation before cap" — the one place that allocates a new string). Also DRY'd `schemaText`'s raw-bytes handling (Gemini).

Declined: hiddenRunes map → switch (Gemini) — the map is O(1) and clear; the RE2 scan dominates, so it's not worth the churn.

Replaced the inconclusive tests with a full-scan no-evasion test (64 KiB+ padding then tail poison → reaches the detector verbatim). Re-validated over SSE: poisoned target still **2 VULN**, DVMCP #9 still **0**. `go test -race`, `go vet`, `golangci-lint` v2.12.2 all clean.

internal/probes/mcptool/toolpoisoning.go
```diff
@@ -0,0 +1,300 @@
...
+	collectStrings(decoded, &sb)
+	return sb.String()
+}
+
+// collectStrings walks a decoded JSON value, appending every map key and string
+// value (recursively) to sb, newline-separated. It strictly bounds the total to
+// maxScanLen — appending only what fits before writing — so an adversarially huge
+// schema (a giant enum value or key) cannot drive a large allocation. Schemas are
+// tiny in practice; this bound only bites on hostile input.
+func collectStrings(v any, sb *strings.Builder) {
```
maurodelossantos-praetorian, line 291: Re-review request for `4191d4f`: this round removed the buggy 64 KiB cap + inconclusive path (they were net-negative — an O(N²)/bypassable `capScan` and a 0.5 score that rendered SAFE) in favor of full-text scanning; only schema extraction stays bounded here in `collectStrings` via per-write capping. Could you take another pass and confirm no critical issues remain — evasions, false positives, false negatives, or correctness gaps — so we can assess merge-readiness? @codex review @gemini review @claude please re-review

maurodelossantos-praetorian:
Sixth round addressed in `ed90865`.

- **Recon empty-inventory silent FN (Codex, critical):** the probe short-circuited whenever *any* recon.MCP inventory existed, even one with no scannable metadata (recon is best-effort and can emit an empty/partial inventory when catalog calls fail). It now falls through to a live `ListTools` enumeration (with a warning) instead of a silent SAFE.
- **Schema bound removed — resolves Gemini's two "critical"s (map-newline cap bypass, unbounded `schemaText` fallback) and Codex's schema-tail evasion together:** the 64 KiB cap was a half-measure that leaked past the guard on wide maps and the unmarshal-failure path, *and* silently dropped tail poison. `collectStrings` now extracts the full schema, consistent with how descriptions are already scanned — the metadata is already resident (recon fetched it) and RE2 is linear-time, so full scanning is O(n) with no evasion. Bounding metadata *ingestion* is a recon/generator-layer concern, not this probe's (worth a separate ticket if we want an ingestion cap across all MCP consumers).
- **Two-step regex FP (Gemini):** the `you must` arm is constrained to action verbs, so "before calling this tool, you must authenticate" no longer fires.

Added a full-schema no-truncation-evasion test and a before-you-must-authenticate FP guard. Re-validated over SSE: poisoned target still **2 VULN**, DVMCP #9 still **0**. `go test -race`, `go vet`, `golangci-lint` all clean.

---

**Meta-observation for the humans:** across six automated rounds the *core* detection has been stable and correct the whole time (2 VULN / 0 FP on live targets since round 1). The allocation-bound saga specifically has now oscillated — round 4 added a cap, round 5 removed part of it, round 6 removed the rest as net-negative — which is the signal that this area has converged on "full scan, ingestion limits belong elsewhere." Remaining findings each round have been progressively narrower. I'd treat this as **ready for human merge review**; further automated rounds are likely to keep surfacing diminishing-returns edge cases rather than correctness defects.

internal/probes/mcptool/toolpoisoning.go
```diff
@@ -0,0 +1,282 @@
...
+func (p *ToolPoisoning) GetPrompts() []string {
+	return []string{"static scan of MCP tool descriptions, input schemas, server instructions, and resource/prompt metadata for hidden LLM-directed instructions"}
+}
+
+// Probe scans the target's advertised metadata and emits one attempt per scanned
+// item (the item's text is the attempt output; the toolsec.ToolPoisoning detector
+// scores it). It prefers a shared recon.MCP inventory (which also exposes server
+// instructions and resource/prompt metadata) and falls back to a live
+// ToolInvoker.ListTools enumeration.
+func (p *ToolPoisoning) Probe(ctx context.Context, gen types.Generator) ([]*attempt.Attempt, error) {
```
maurodelossantos-praetorian, line 75: Re-review request for `ed90865` — aiming to confirm merge-readiness. This round: fixed the recon empty-inventory silent FN (this Probe method now falls through to a live ListTools enumeration instead of returning zero attempts), removed the net-negative schema size cap in favor of full-text scanning (metadata is already resident from recon; RE2 is linear-time; bounding ingestion is a recon-layer concern), and constrained the two-step 'you must' regex to action verbs to kill a benign-orchestration false positive. Please take another pass and flag any remaining CRITICAL correctness issues — evasions, false positives, false negatives — that should block merge. @codex review @gemini review @claude please re-review

maurodelossantos-praetorian:
Rebased onto latest `main` by merging `origin/main` (`0ee88a2`). `main` renamed the `toolsec` package to **`mcptool`**, so this probe/detector are now **`mcptool.ToolPoisoning`** (metadata keys `mcptool.tool` / `mcptool.poison_location` / `mcptool.poison_signal`). Moved the files into `internal/{probes,detectors}/mcptool/`, folded the package-doc note into `mcptool.go`, and regenerated `pkg/register` (`make generate-check` passes). Conflicts resolved; PR is MERGEABLE. Re-validated over SSE: poisoned target **2 VULN**, DVMCP #9 **0**. build / go test -race / lint / generate-check all green. (Note: earlier comments referencing `toolsec.ToolPoisoning` are now `mcptool.ToolPoisoning`.)

internal/probes/mcptool/toolpoisoning.go
```diff
@@ -0,0 +1,282 @@
...
+func (p *ToolPoisoning) GetPrompts() []string {
+	return []string{"static scan of MCP tool descriptions, input schemas, server instructions, and resource/prompt metadata for hidden LLM-directed instructions"}
+}
+
+// Probe scans the target's advertised metadata and emits one attempt per scanned
+// item (the item's text is the attempt output; the mcptool.ToolPoisoning detector
+// scores it). It prefers a shared recon.MCP inventory (which also exposes server
+// instructions and resource/prompt metadata) and falls back to a live
+// ToolInvoker.ListTools enumeration.
+func (p *ToolPoisoning) Probe(ctx context.Context, gen types.Generator) ([]*attempt.Attempt, error) {
```
maurodelossantos-praetorian, line 75: Post-rebase review request (`0ee88a2`, now on latest main; package renamed toolsec→mcptool, so this is `mcptool.ToolPoisoning`). This `Probe` reads the target's advertised metadata (via a recon.MCP inventory or a live ListTools fallback) and emits per-item attempts scanned in full. Please re-review the whole probe/detector after the merge and flag any CRITICAL or MAJOR issues (evasions, false positives/negatives, correctness) before merge. @codex review @gemini review @claude please re-review @coderabbitai review

maurodelossantos-praetorian:
## Ready for human merge review ✅

Rebased onto latest `main` (package `toolsec`→`mcptool`; this is now **`mcptool.ToolPoisoning`**), conflicts resolved, and the post-rebase review round addressed. All checks green: `build`, `go test -race`, `golangci-lint`, `make generate-check`.

**What this PR does:** adds server-integrated **tool-poisoning** detection (OWASP MCP03 + MCP10) — a deterministic, no-LLM probe/detector that reads the target's own advertised metadata (tool/resource/prompt descriptions, titles, schemas, server instructions) via a `recon.MCP` inventory or a live `ListTools` fallback and flags hidden LLM-directed instructions (smuggling tags, secrecy/exfiltration/shadowing directives, invisible-Unicode).

**Addressed in the latest commit:**
- **CodeRabbit (MAJOR):** catalog pagination — recon and `ListTools` now follow `nextCursor` across all pages (bounded, cursor-repeat-guarded), so poison hidden on a later page can't evade the scan.
- **Codex P1:** on a partial recon (tools/list failed, only non-tool metadata present) the probe now falls through to a live `ListTools` scan instead of short-circuiting.
- **Codex P1:** the live path's SDK schema object is now marshalled before scanning, so schema-embedded poison is caught on the no-recon path too.

**Validation:** re-verified over SSE — poisoned target **2 VULN**, DVMCP #9 (benign) **0**, on both the recon and no-recon paths; added tests for pagination (+cursor-repeat), non-map schema scanning, and partial-recon merge. Earlier rounds (Claude, Codex, Gemini) were LGTM/no-critical.

maurodelossantos-praetorian:
## Follow-up review addressed (`bf1e6e3`)

Re-assessed the latest agent comments on the PR:

**CodeRabbit (Major) — silent page-cap truncation in `listAll`:** valid, fixed. `listAll` used to return success after `maxListPages` (1000) even when a fresh cursor was still pending, so a hostile catalog with >1000 unique-cursor pages could hide poison past the cap and be reported as a complete, clean scan. Now:
- `listAll` returns a sentinel `errListTruncated` (with the items gathered so far) on cap exhaustion — never a silent partial-as-complete.
- Recon keeps the partial catalog but **logs a warning** on truncation (a real list error still leaves that catalog empty, as before).
- `mcp.MCP.ListTools` keeps the partial list + warns rather than hard-failing the whole scan.
- Added a regression test asserting `errListTruncated` (and that the gathered items are retained) on cap exhaustion.

**Earlier Codex P1s (recon-partial fall-through, live-schema marshalling) and the CodeRabbit MAJOR pagination finding:** already addressed in the previous commit `f76fd4b` — no further action needed.

**Nothing outstanding.** Re-validated over SSE: poisoned target **2 VULN**, DVMCP #9 (benign) **0**. `build`, `go test -race`, `golangci-lint`, `make generate-check` all green. PR is `MERGEABLE`.

### Conclusion
All agent findings across every round are now resolved or confirmed non-issues; the core detection has been correct and stable throughout, and the recent rounds hardened pagination, partial-recon handling, and truncation transparency. This is good to go for human merge review.

maurodelossantos-praetorian:
## Pausing & closing — deprioritizing malicious-server *detection*

Closing this PR. This is a **strategic pause, not a quality issue** — the probe is complete, tested, lint/CI-green, and validated live against a poisoned SSE target. The branch and full review history are preserved, so this is fully recoverable (reopen anytime).

**What it does:** `mcptool.ToolPoisoning` statically inspects an MCP server's advertised tool/resource metadata and flags hidden LLM-directed instructions (OWASP MCP03 / MCP10 tool poisoning).

**Why we're pausing it:** it *detects whether a server is malicious/poisoned* rather than *finding an exploitable vulnerability in the server*. That's detection, not attack — a different question ("is this server trustworthy?") for a different beneficiary (the agent's owner) than Augustus's core "scan a target, find its vulnerabilities" mission. As a server-vuln scanner it doesn't earn a slot, and detecting malicious/vulnerable servers (this, and the earlier `RugPull`/`ToolDrift`) isn't a priority right now.

**What would earn a slot later:** the *dynamic* half — present the poisoned tool defs to a host LLM and demonstrate the agent **obeys** them (an actual attack with impact), and/or reframing the static scan as a recon input feeding that attack. Captured on LAB-4460.

**Not lost:** the CodeRabbit-flagged **pagination fix** (`recon.go` / `mcp.go` now follow `nextCursor` across all pages) is generator-layer and benefits *every* MCP probe — it'll be salvaged into its own PR rather than lost here.

LAB-4460 is being moved back to the backlog.

maurodelossantos-praetorian:
Closing as deprioritized (strategic — malicious-server *detection* isn't the current priority; see the detailed comment above). Branch preserved; reopenable. Pagination fix to be salvaged separately.

PR #263, opened 2026-07-22
feat(recon): callable recon orchestration + rename toolsec.* -> mcptool.*

praetorian-farida:
Two related changes so Guard (and other embedders) can drive the recon phase and find the MCP tool-security probes consistently.

## 1. Callable recon orchestration (`pkg/recon/orchestrate.go`)
Extracts the recon create+run+context-inject sequence out of `cmd/augustus/scan.go` into reusable helpers so external consumers don't duplicate the CLI wiring:
- `CreateAll(names, resolve) ([]Recon, error)` — best-effort build; skips and joins per-module construct failures.
- `RunAll(ctx, gen, names, resolve, store) error` — create + run in one call.
- `InjectProbeContext[T any](items, store)` — generic, no probes import.
- `ConfigResolver func(name string) registry.Config`.

`cmd/augustus/scan.go` now calls these and drops its private `createRecons`/`injectProbeContext`. Side benefit: the CLI was abort-on-first-construct-failure; it's now best-effort.

## 2. Rename `toolsec.* -> mcptool.*`
Renames the MCP tool-surface probe/detector family from `toolsec` to `mcptool` so it sits alongside the existing `mcpconfig.*` and `mcptransport.*` families (config / transport / tool-surface) and is easy to find. Renames both `internal/probes/toolsec` and `internal/detectors/toolsec` packages and every `toolsec.*` registration/name; probe<->detector wiring stays aligned. No behavior change.

Full test suite green.


blayne:
# PR #263 Review — `feat(recon): callable recon orchestration + rename toolsec.* → mcptool.*`

**Recommendation: 🔴 REQUEST CHANGES** — 1 medium blocker (fix is small).

No Linear ticket linked, so this is a code/security/test/docs review without acceptance-criteria alignment. Verified locally: `go build ./...` clean, `go vet` clean, zero residual `toolsec` identifiers, `CLAUDE.md` fully migrated.

The rename half of the PR is clean and mechanical — registrations, probe↔detector wiring, and docs all migrate consistently, no behavior change. The one substantive issue is a side effect of the abort→best-effort change, independently flagged by both the security and quality reviewers.

---

## 🔴 SEC-BE-001 (medium, blocker) — fail-open on recon construction failure for full scans

`cmd/augustus/scan.go:715-726`

The refactor changes recon **module-construction** failures from a hard scan abort into a log-only condition on the recon+probes path. `recon.RunAll` joins construct + run errors into `reconErr`, but `reconErr` is consumed **only** in the recon-only branch (`scan.go:782-783`, which correctly fails loud). On a full scan it's set at line 725 and never read again — the scan proceeds and the exit code is unaffected.

Pre-PR, `createRecons` returned its error and the caller did `return rerr`, so a construction failure aborted the whole scan.

Why it matters: `mcptool.BOLA` returns `nil, nil` on an empty identifiers store (`internal/probes/mcptool/bola.go:84-85`) and can only get identifiers from `recon.MCPIdentifiers`. So if that module fails to construct — a typo'd `--recon` name, a missing/misconfigured judge generator, or a malformed `recon.settings` block — BOLA silently produces zero attempts, the report shows no BOLA findings, and the process exits 0. An operator reads that as "not vulnerable to cross-tenant/IDOR reads" when BOLA never actually ran. Only a single `slog.Warn` marks the degradation. (SSRF/Injection/PathTraversal fall back to a live `ListTools`, so only BOLA hard-depends on recon and is exposed.)

The recon-only path already documents a "do not report a false green" principle (`scan.go:776-779`); the probe path just doesn't apply it.

**Fix options:**
- Extend the recon-only guard to the probe path: when recon modules were requested and `reconErr != nil` (at minimum for construction/registration errors — those are operator misconfig, not partial-data), fail the scan instead of only warning; **or**
- Have recon-dependent probes distinguish "context absent" from "context present but empty" and emit an inconclusive (`0.5`) attempt, so a missing prerequisite surfaces as a finding a reviewer must clear rather than a silent green.

Either way, the exit code and report — not just the log — should reflect that a requested security-recon module did not run.

---

## 🟢 TEST-001 (low, non-blocking) — informational coverage note

`pkg/recon/orchestrate_test.go`

New helpers have 100% statement coverage. The one combination not isolated in a `pkg/recon` unit test — a module that constructs fine but errors during `Run` — is covered transitively by `cmd/augustus` `TestScanCommand_ReconOnly_FailsOnModuleError` and by `TestRun_StampsSourceAndAggregatesErrors`. Not a functional gap; a dedicated symmetry test is optional.

---

## Cleared

- **API design** of `CreateAll` / `RunAll` / `InjectProbeContext[T]` / `ConfigResolver` — accurate, thorough doc comments.
- **`errors.Join` semantics** — construction errors survive even when zero modules build; `RunAll` calling `Run([])` is harmless (`errors.Join()` == nil).
- **nil-safety** — `CreateAll` guards `if resolve != nil`; nil result → nil cfg, which every recon factory reads safely; the `resolve` closure over a possibly-nil `yamlCfg` is safe (`ResolveReconConfig` has a nil-receiver guard).
- **Generic `InjectProbeContext[T any]`** — `any(it).(ContextAwareProbe)` is equivalent to the old concrete assertion; before-buff-wrap contract preserved.
- **Rename** — zero residual `toolsec`; registrations, `mcptool.tool`/`mcptool.param` metadata handshake, and both `pkg/register` files consistent; injection/path/SSRF logic byte-identical aside from identifiers.
- **Docs** — `CLAUDE.md` fully migrated.

_Reviewed at head `39679f1`. Prune/edit before posting._

blayne (changes requested):
See PR comment.

praetorian-farida:
@blayne thanks for the thorough review — both flagged items are addressed in `7b91c77`.

## SEC-BE-001 (blocker) — fixed

I took a third approach rather than either of your two options, and want to flag the divergence on purpose:

Instead of failing before the probe phase on construction errors (option 1) or emitting an inconclusive `0.5` from recon-dependent probes (option 2), the scan now **folds `reconErr` into the exit status after `harness.Run`** (`cmd/augustus/scan.go`). This keeps recon fully best-effort — surviving modules run, their observations still emit, and the probe phase still runs on whatever was gathered — but **any** recon failure (construction *or* runtime) now surfaces as a non-zero exit.

Two reasons for going broader than "construction errors only":
- A module that builds fine but errors at runtime and gathers nothing produces the same BOLA-no-op false green as a construction failure, so gating on construction alone left a residual hole.
- It keeps the full-scan path symmetric with the recon-only path's existing "no false green" guard, without discarding probe results on a partial recon hiccup.

Net effect matches your principle: the exit code and report — not just the log — reflect that a requested recon module didn't complete.

## TEST-001 (non-blocking) — covered

The construct-fine-but-errors-at-runtime path is now exercised directly at both levels:
- `TestRun_StampsSourceAndAggregatesErrors` (pkg/recon) — survivors' observations preserved + error aggregated
- `TestScanCommand_FullScan_FailsLoudOnReconRunError` (cmd) — full scan exits non-zero on a recon run error while probes still run

Plus `TestScanCommand_FullScan_FailsLoudOnReconConstructionError` for the construction case.

Build / `go vet` / `golangci-lint` clean. The earlier CodeRabbit nits (import order, BOLA log wording, OriginValidation comment) were fixed in `39679f1`.

One sharp edge left as-is by design: a module returning *partial* observations *alongside* an error has its own partial gatherings dropped by `Run` (`continue` before `store.Observe`). Pre-existing, affects only the erroring module, out of scope here — happy to file a follow-up if you think it's worth changing.

blayne (approved):

PR #265, opened 2026-07-24
feat(probes): expose curated RiskInfo write-ups on MCP probes

praetorian-farida:
## What

Adds an optional `types.RiskDescriber` interface + `types.RiskInfo` struct so a probe can carry its own security write-up (vulnerability description, impact, remediation, references, CWE taxonomy, and a conservatively-scored CVSS v4.0 vector), and implements `RiskInfo()` on the eight MCP probes:

- `mcptool.Injection` / `SSRF` / `BOLA` / `PathTraversal` / `ResponseLeak`
- `mcptransport.OriginValidation` / `SSESessionHijack`
- `mcpconfig.CredentialExposure`

## Why

The finding write-ups for these probes currently live in a hand-maintained map in the consumer (Guard). Per review feedback, the write-up should live next to the probe so consumers retrieve it (via a `types.RiskDescriber` type assertion) instead of duplicating and drifting. This keeps the security content with the logic it describes and lets Guard/Hadrian/etc. render consistent definitions.

## Notes

- Optional interface, mirroring the existing `ProbeMetadata` pattern — probes that don't implement it are unaffected; consumers fall back to their generic path.
- CVSS vectors are scored **conservatively** (no scope change, tool invocation requires client access `PR:L`, partial/conditional impact as `L`, honest preconditions like `UI:A`/`AT:P`) — a real spread (~5.3 Medium .. 8.7 High) rather than a flat High/Critical cluster.
- Compile-time assertions + per-package tests guard interface conformance and completeness.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -42,6 +42,7 @@ Probes may also implement these **optional** interfaces (all in `pkg/types/probe
 - **ProbeDetectorConfig**: `GetDetectorConfig()` — per-probe detector config overrides
 - **ProbeSecondaryDetectors**: `GetSecondaryDetectors()` — run extra detectors alongside the primary; the attempt verdict reflects the **max score across all detectors** (see `attempt.GetEffectiveScores`), so a secondary hit alone marks the attempt vulnerable
 - **ProbeTools**: `GetTools()` / `GetToolChoice()` — declare function-calling tool schemas for tool-use/agent probes (sent via the native wire layer in `internal/attackengine/toolcalls.go`)
+- **RiskDescriber**: `RiskInfo()` — carry a curated `types.RiskInfo` write-up (description/impact/recommendation/references/taxonomies/CVSS v4.0 vector) next to the probe so consumers retrieve it via type assertion instead of duplicating it
 
 Generators may also implement these **optional** interfaces (in `pkg/types/generator.go`):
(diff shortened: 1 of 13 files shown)


praetorian-farida:
Addressed the actionable bot feedback in 469f4c8:

- **Codex (critical)** — `SSESessionHijack.RiskInfo` no longer claims sampled IDs showed limited entropy/shared prefixes, and CWE-330 is dropped from references + taxonomies. The write-up now covers only what the probe tests (connection/lifetime binding, CWE-287/CWE-613).
- **Drift check** — documented the new `RiskDescriber` optional interface in CLAUDE.md.
- **Gemini (CVSS validation)** — added a `CVSS:4.0/` prefix guard to the `riskinfo_test` completeness checks.

Two suggestions intentionally **not** taken:

1. **Gemini — change `References`/`Taxonomies` to `[]string`.** These are modeled as newline-separated strings deliberately, to match the format consumers already render taxonomy/reference blocks in (the same `- cwe: N` block shape used elsewhere). Switching to `[]string` would change the public `types.RiskInfo` shape and push a formatting decision that currently lives in one place out to every consumer, for no correctness gain. Leaving as-is; can revisit if we standardize the consumer-side rendering.

2. **Claude — `CVSSVector` doc says "optional" but tests require it.** No real contradiction: the field is optional at the interface/type level (a future probe may implement `RiskDescriber` without a curated severity), while the per-package `riskinfo_test` completeness checks enforce it for these 8 concrete probes because they all do have a curated vector. The type doc and the probe-specific tests are describing two different contracts, so `prober.go` is left unchanged.

PR #274, opened 2026-07-30
fix(mcp): follow nextCursor across all catalog pages (LAB-5094)

maurodelossantos-praetorian:
Closes LAB-5094.

## Problem

Every MCP catalog listing in Augustus asked for **page one and stopped**. The MCP spec paginates `tools/list`, `resources/list`, `resources/templates/list` and `prompts/list` via an opaque `nextCursor`, but all our call sites passed `nil` params. The go-sdk doesn't cover for us either — `ClientSession.ListTools` is a single request/response; the SDK's paginating iterator (`ClientSession.Tools`) is a separate API we never called.

Against a server that paginates its catalog, `mcptool.Injection`, `.SSRF`, `.PathTraversal`, `.BOLA` and `.ResponseLeak` would test only the first page of tools and report the target **clean**. The untested tools appeared nowhere in the output, so nothing signalled that coverage was partial — a silent false negative, which is the worst failure mode a scanner can have. Low probability (most catalogs fit one page), bad consequence.

Fixing it at the generator/recon layer fixes it once for every current and future MCP probe.

## Change

A bounded generic `listAll` helper walks `nextCursor`, accumulating items and stopping on an empty cursor, a **repeated** cursor, or a hard page cap (1000). On cap exhaustion it returns `errListTruncated` *alongside the items gathered so far*, so a truncated enumeration is never reported as complete — the hostile-server case, not just the benign one. A `reconList` wrapper adapts it to the inventory's existing best-effort style (warn and keep on truncation, nil on any other error, so a catalog that fails to enumerate stays empty exactly as before).

Applied to all three catalog-consuming paths:

| Path | Who depends on it |
|---|---|
| `MCPInventory` — tools, resources, resource templates, prompts | `recon.MCP`, and every context-aware probe that reuses the inventory |
| `ListTools` (`types.ToolInvoker`) | every `mcptool.*` probe |
| `listTools` (`mode: list_tools`) | detectors inspecting advertised names/descriptions/schemas |

## Notes for review

- **The third call site was easy to miss.** The original fix (written on the now-closed #262) covered `MCPInventory` and `ListTools` but not `listTools`, the `mode: list_tools` rendering path, which had the identical bug. All three are covered here.
- **`LastRawResponse` semantics.** It keeps its documented "most recent" meaning, now holding the most recent *page*. Concatenating pages would not be valid JSON, and behaviour is unchanged for single-page servers.
- **Salvage provenance.** The `listAll`/`reconList` core and the unit tests come from the closed #262 (`lab-4460`), cherry-picked file-by-file rather than by merging the branch — that branch predates #265/#273 and merging it would have reverted the `RiskInfo` work. An unrelated `title` projection change from that branch was deliberately left out to keep this PR to one concern.

## Testing

Unit tests cover `listAll` in isolation: cursor following across pages, cursor-repeat detection, error propagation, and `errListTruncated` on cap exhaustion.

The end-to-end tests are the guard that matters. They stand up a real SDK server over streamable HTTP with `ServerOptions.PageSize: 2` and 5 items per catalog, so the catalogs genuinely span three pages, and drive all three paths through the actual wire protocol. **Each one fails against a single-page client, returning 2 of 5 tools** — verified by neutering pagination and re-running:

```
--- FAIL: TestToolInvoker_ListTools_EnumeratesEveryPage
    ListTools returned 2 tools, want 5 (later pages dropped)
--- FAIL: TestMCPInventory_EnumeratesEveryPage
    inventory tools = 2, want 5 (later pages dropped)
    inventory resources = 2, want 5 (later pages dropped)
    inventory resource templates = 2, want 5 (later pages dropped)
    inventory prompts = 2, want 5 (later pages dropped)
--- FAIL: TestGenerate_ListToolsMode_EnumeratesEveryPage
```

This also closes a standing gap noted while scoping the work: the `mcptool` probes had no end-to-end coverage over the real MCP wire protocol, which is precisely why this bug shipped without a failing test.

Verification run locally:

- `go test ./internal/generators/mcp/... ./internal/probes/mcptool/... ./internal/probes/mcptransport/... ./internal/probes/mcpconfig/... ./internal/recon/... -race` — all pass
- `golangci-lint fmt ./...` (v2.12.2, the CI-pinned version) — no drift
- `golangci-lint run ./internal/generators/mcp/...` — 0 issues
- `go build ./...`, `go vet` — clean


maurodelossantos-praetorian:
## AI review triage — 6 true positives fixed, 2 declined

Thanks to all four reviewers; this was a genuinely productive pass. Every finding was checked against the current code rather than taken at face value. Fixed in `514d895`.

### True positives — fixed

**1. Cursor repetition was reported as a complete enumeration** *(Codex — critical; CodeRabbit; Claude — noted)*

Confirmed and the most serious of the set. `listAll` returned `nil` on `seen[next]`, so a hostile server could halt the walk after page one by replaying a cursor and have the partial catalog reported as the target's **full attack surface** — no warning, nothing marking it incomplete. That directly contradicted the helper's own "never partial-as-complete" contract, which the page cap and walk budget both honoured. Three reviewers flagged it; Claude called it defensible-but-worth-a-comment, and I disagree — in a scanner the asymmetry is a false-negative vector, not a style question. Only an empty `nextCursor` now means the catalog ended. `TestListAll_StopsOnCursorRepeat` updated accordingly, as CodeRabbit correctly anticipated.

**2. One walk budget shared across all four catalogs** *(Codex — security)*

Confirmed, and my own design error. A large or deliberately slow `tools/list` could exhaust the shared budget and force resources, resource templates, and prompts to each return **empty** — hiding hostile resource and prompt definitions behind a slow tool list, which is precisely the failure the bound exists to prevent. The budget is now derived per catalog inside `reconList`.

**3. `LastRawResponse` held only the final page** *(Gemini — critical)*

Confirmed, and this one corrects my earlier reasoning. I had defended per-page storage on the grounds that a combined catalog "wouldn't be valid JSON" — but Gemini's suggestion isn't concatenation, it's a single `ListToolsResult` carrying every tool, which is both valid and strictly more faithful. It matters concretely: `pkg/hooks/generator.go` writes this to a temp file for runtime hooks, so a hook scanning a paginated catalog saw one tool where the target advertised several. Rebuilt from the last page so its `_meta` and a still-pending `nextCursor` survive.

**4. Caller cancellation was masked as truncation** *(Gemini — suggestion)*

Confirmed. `walkCtx.Err()` conflated *our* budget expiring with *the caller* tearing the scan down, so a `Ctrl-C` or `--timeout` mid-walk logged "results may be incomplete" and carried on against a catalog it had been told to stop building. The parent context is now checked first and propagates; only the walk budget maps to truncation.

**5. A truncated catalog was memoized as canonical** *(CodeRabbit — actionable)*

Confirmed. `toolsCache` lives for the session, so an incomplete tool surface was handed to every later probe with no further signal. The truncated set is still served to the current caller, but not cached — one extra `tools/list` buys the next caller a full walk.

**6. The paginated `tools/list` walk was written three times** *(CodeRabbit — nitpick)*

Taken, and it earned its keep: with four behavioural fixes landing in that same block, three copies were a real divergence risk. Now one `listAllTools` helper. **Recon keeps its own closure deliberately** — it drives four catalogs through one generic helper and skips `storeRawJSON`, so folding it in needs a flag and would make `tools` inconsistent with its three siblings.

Also taken from CodeRabbit: `reconList` now takes `ctx` first, and logs hard enumeration failures instead of returning `nil` silently.

### Test feedback — taken with one deliberate exception

`t.Parallel()` on both timing tests takes the package from **~13s to ~5.5s**. The walk-bound test's elapsed-time assertion is replaced by its page-count assertion, which is the stronger signal exactly as CodeRabbit argued: server-side delays mean an unbounded walk *must* serve all 40 pages, so the check is exact and machine-independent, where the 7s threshold overlapped the 5s bound too closely to survive a loaded or parallel box.

**Declined: gating the walk-bound test behind `testing.Short()`.** Earlier in this branch a `t.Skip` in this very test silently hid the regression it existed to catch — it passed with the bound removed. A skip that can mask a missing security bound is worse than 5 seconds of wall clock, so these assertions stay unconditional.

### Not a defect

`reconList`'s parameter ordering was raised as optional; CodeRabbit correctly noted the context-as-argument rule isn't a gated linter here. Done anyway since it's idiomatic, but it wasn't a correctness issue.

### Verification

Five new tests, and **each was verified to fail against the behaviour it replaces** rather than merely passing on the new code:

| Fix | Neutered behaviour | Result |
|---|---|---|
| Cursor repeat | returns `nil` again | FAIL |
| Cancellation | parent `ctx.Err()` check removed | FAIL |
| Per-catalog budget | one shared `walkCtx` | FAIL (resources/prompts empty) |
| Raw response | store final page only | FAIL (`missing "tool-00"`…) |
| No-memoize | cache regardless of truncation | FAIL (`20 of 40 tools was memoized`) |

`go test ./... -count=1` clean; `-race` clean across all MCP and recon packages; `golangci-lint fmt` no drift; `golangci-lint run` 0 issues (v2.12.2, CI-pinned).

maurodelossantos-praetorian:
## Follow-up review finding — also a true positive, fixed in `dbf514f`

**`listTools` dropped the truncation signal reaching detectors** *(CodeRabbit)*

Confirmed. `listAllTools` computes whether the walk was truncated and `listTools` discarded it with `_`, so a `mode: list_tools` run rendered a partial catalog with nothing marking it partial — only an operational `slog.Warn`.

This matters more than its "trivial" label suggests. That rendered text **is** the artifact a detector scores, so without the notice a detector judges an incomplete tool surface as though it were the target's whole one and reports clean. That's the same silent false negative this branch exists to remove, one layer up. It's also precisely the log-only limitation I'd flagged as a known gap on the ticket, now closed at the one path that has a channel for it.

### On the mechanism

I checked whether a metadata key would be more appropriate than appending to the message text, since mixing scanner-generated text into what is nominally target output is normally an anti-pattern — it risks a detector scoring our own words as the target's. **`attempt.Message` has no metadata field** (only `Role`, `Content`, `Images`, `Audio`, `Documents`, `ToolCalls`, `ToolCallID`), and the established `MetadataKey*` constants are set by probes, not reachable from a generator. So for this path the rendered text is genuinely the only channel, and a metadata key would mean widening the generator/attempt API — out of scope here.

Given that, the notice is worded to mitigate the anti-pattern rather than ignore it: clearly attributed to `augustus`, no imperative or instruction-like phrasing that a detector scanning for injected instructions could mistake for target output, and appended **only** on truncation so a normal walk's output is byte-identical.

### Scope note

The `ToolInvoker` path deliberately keeps log-plus-no-memoize instead. Its result is `[]map[string]any`, where the only way to inject a signal would be a synthetic entry indistinguishable from a real tool — strictly worse than staying quiet. Not caching the truncated catalog already gives the next caller a full walk.

### Verification

`TestGenerate_ListToolsMode_MarksTruncatedCatalog` asserts **both** directions: the notice appears on a truncated walk, and is absent on a complete one, so scanner metadata only enters detector input when it's actually true. Verified to fail when the flag is discarded again.

Full suite, `-race`, `golangci-lint fmt` and `run` all clean.

---

**Running tally for this PR: 11 defects found, 0 by CI.** Five by my own adversarial reading of the diff, six from you four. CI was fully green on the commit that would have let a looping server pass a one-page catalog off as a complete attack surface — so the review pass earned its place here. Thanks.

internal/generators/mcp/recon.go
```diff
@@ -196,3 +226,131 @@ func mcpPromptsFrom(prompts []*mcpsdk.Prompt) []types.MCPPrompt {
...
+		// Reporting it as such would have the scan carry on against a partial catalog
+		// it was told to stop building, so propagate it and let the caller unwind.
+		// Checked before walkCtx because walkCtx is derived from ctx and would
+		// otherwise mask a cancellation as our own budget expiring.
+		if err := ctx.Err(); err != nil {
+			return out, err
+		}
+		if walkCtx.Err() != nil {
+			return out, errListTruncated
+		}
```
maurodelossantos-praetorian, line 385: @claude @codex review @gemini — re-review request on `dbf514f`.

Your previous pass reviewed **`83fb618`** (Codex confirmed this explicitly: *"Reviewed commit: 83fb618002"*). Two commits have landed since, and they contain most of the interesting logic — so this code has not actually been reviewed yet:

- **`514d895`** — fixes six findings from your last pass: cursor repetition now reports truncation instead of completion; the walk budget is per-catalog rather than shared across all four; `LastRawResponse` carries the whole catalog instead of the final page; caller cancellation propagates instead of being masked as truncation; truncated catalogs are no longer memoized; the triplicated `tools/list` walk is now one `listAllTools` helper.
- **`dbf514f`** — surfaces truncation in the probe-facing `mode:list_tools` output, not just the operator log.

**Please focus on the two invariants this anchor line sits on**, since they are new and subtle:

1. **The three-way distinction right here.** Caller cancellation (line 306) must propagate; our own walk budget expiring (line 309) must degrade to `errListTruncated`; an empty `nextCursor` (line 317) is the *only* clean completion. Is that separation actually airtight? Specifically: `walkCtx` is derived from `ctx`, so is checking `ctx.Err()` first genuinely sufficient to stop a caller cancellation being misreported as truncation — including when the parent carries its own deadline via `--timeout`, where both would surface as `DeadlineExceeded`?

2. **Whether `errListTruncated` is now handled consistently at every consumer.** It is produced in three places (page cap, cursor repeat, walk budget) and consumed in four (`reconList`, `listAllTools`, `ListTools`, `listTools`). A consumer that treats it as a hard error, or that ignores it, reintroduces exactly the silent-under-scan bug this PR exists to fix.

Also worth adversarial attention: the per-catalog `walkBudget()` in `MCPInventory` is now computed once and passed to four independent `reconList` calls — please confirm each really gets its own fresh deadline and there is no residual sharing, since that was Codex's security finding last round and I do not want to have half-fixed it.

Context on scale: this PR has had **11 defects found in it, none by CI** — five from my own review of the diff, six from yours. So please be sceptical rather than confirmatory; the automated gates were fully green on the commit that would have let a looping server pass a one-page catalog off as a complete attack surface.

maurodelossantos-praetorian:
## Second review round — 4 more true positives, fixed in `e741bb2`

All four are real. Two were security-relevant, and one of them is the most important finding this PR has had. Notably, **one suggested diff would have reintroduced a bug I fixed earlier in this branch** — details below.

### 1. `ListTools` returned a knowingly partial catalog with a nil error *(Codex — critical)*

Confirmed, and this goes to the PR's central claim. It reproduced the very defect this branch exists to remove, one layer up: a server can halt the walk after a benign prefix, and every `mcptool.*` probe would then scan that prefix, find nothing, and report the target **clean**.

`ListTools` now **fails closed** — an honest "could not enumerate the surface" beats a false pass. The partial catalog still travels alongside the error, so a caller that deliberately wants best-effort data can use it, while the default `if err != nil` path skips the target.

Note the deliberate asymmetry with `mode:list_tools`, which keeps returning its catalog plus the truncation notice: that path renders **text a detector scores**, so incompleteness can be expressed in-band. `ListTools` returns `[]map[string]any`, where the only way to signal in-band would be a synthetic entry indistinguishable from a real tool. So: express incompleteness in-band where a channel exists; fail closed where none does.

### 2. Truncated inventories had no completeness marker *(Codex — security)*

Confirmed, and **the consumer half mattered more than the marker**. `resolveTools` in `internal/probes/mcptool/recon_context.go` had:

```go
if len(tools) > 0 { return tools, nil }
```

— exactly the "prefers any non-empty inventory" path you described. A hostile server that stopped recon after a benign prefix had every context-aware probe (`Injection`, `SSRF`, `BOLA`, `PathTraversal`, `ResponseLeak`) reuse it as the full surface.

`types.MCPInventory` now carries `Incomplete []string` plus `IsComplete()`, naming precisely which catalogs stopped early, and `resolveTools` ignores an incomplete inventory and re-enumerates live — which itself fails closed per #1. The marker is `omitempty`, so a complete inventory serializes byte-identically to before (asserted by a test). The `pkg/types` change went through the repo's AGENT-FIRST path.

### 3. Page deadlines could overrun the walk budget *(CodeRabbit — major; Codex — suggestion)*

Confirmed: a page starting just before the budget expired could run a further full `request_timeout`, so the advertised bound could be overrun by up to 10%.

**CodeRabbit's suggested diff alone would have reintroduced the empty-catalog bug** that commit `c0d8923` exists to fix. Deriving `pctx` from `walkCtx` means a walk-cancelled page surfaces as a transport error, and `reconList` converts a non-truncation error into `nil` — collapsing the catalog to empty, the outcome strictly worse than truncation. CodeRabbit half-anticipated this ("ensure walk-budget expiry remains `errListTruncated`"), and Codex stated the complete fix outright: *"Derive the page context from both limits and translate walk-budget expiry into truncation."*

So `listAll` now classifies by **which** deadline fired rather than by the error the SDK surfaced: caller cancellation propagates, walk expiry becomes truncation, and the pages already collected survive either way. `listAll` also owns both contexts now, which makes per-catalog budgets **structural** instead of a convention every caller has to remember — that also answers the "confirm each really gets its own fresh deadline" question from my re-review request.

### 4. A non-positive `request_timeout` expires instantly *(Claude — minor)*

Confirmed. `0` means "no timeout" elsewhere in Augustus (`--timeout`, `--probe-timeout`), but here it was applied as an already-expired deadline: every page failed instantly and the scan reported an **empty tool surface** rather than an error. Clamped to the default with a warning, matching how a below-floor rate limit is handled. Fixed despite being flagged non-blocking, because "silently reports an empty surface" is precisely this PR's theme.

### Verification

Five new tests, each verified to fail against the behaviour it replaces:

| Fix | Neutered behaviour | Result |
|---|---|---|
| Fail closed | return `tools, nil` | FAIL |
| Completeness marker | drop `Incomplete` | FAIL |
| Page bounded by walk | `pctx` from caller | FAIL (`9.9s past the walk deadline`) |
| `request_timeout` clamp | take the value literally | FAIL |

**One test of mine did not discriminate on the first attempt** and I want to flag it rather than bury it: the page/walk deadline fix is at most a 10% overshoot, and my initial page-count assertion passed with the fix removed. The test now asserts the deadline a page *actually receives* — deterministic, no sleeps — and fails by 9.9 seconds when the bound is removed.

Full `go test ./...`, `-race` across all MCP/recon/types packages, `golangci-lint fmt` and `run` (v2.12.2, CI-pinned) all clean. Changelog updated for the three user-visible changes.

---

**Tally: 15 defects found in this PR, 0 by CI.** Five from my own review of the diff, ten from you four. Worth flagging for the repo: the Claude/Codex/Gemini workflows trigger on `opened` and `ready_for_review` but **not** on `synchronize`, so pushes don't re-trigger them — which is why `514d895` and `dbf514f` went unreviewed until I asked explicitly. Anyone relying on these reviewers should know a push alone doesn't get a fresh pass.

maurodelossantos-praetorian:
## Third round — 3 more true positives, fixed in `d9a877b`

### 1. `reconList` swallowed caller cancellation *(CodeRabbit — major)*

Confirmed, and this one stings: it **defeated the cancellation fix from the previous commit one layer up.** `listAll` correctly classified a `Ctrl-C` or expired `--timeout` as a cancellation rather than truncation, but `reconList` flattened every non-truncation error into "leave that catalog empty" and `MCPInventory` always returned `nil` — so an aborted scan produced a successful-looking inventory whose empty catalogs are indistinguishable from a target that simply advertises nothing.

This is exactly the *"is `errListTruncated` handled consistently at every consumer?"* risk I asked you to check when requesting this re-review, and I had missed it myself. `reconList` now returns three deliberately distinct outcomes: truncated keeps what it gathered and marks the catalog incomplete; a failed list call still leaves that catalog empty (the best-effort contract exists for a partially reachable **server**); a caller abort returns the error for `MCPInventory` to propagate.

### 2. `resolveTools`'s incomplete-inventory rejection had no test *(CodeRabbit — actionable)*

Fair, and it deserved one — it's the consumer half of the hostile-server path from round two, where a server that halts recon after a benign prefix would otherwise have every probe score that prefix. Added `TestInjection_IgnoresIncompleteReconInventory` as the mirror of the existing `TestInjection_PrefersReconStore`: the stored inventory is marked incomplete and advertises only a benign tool, the live target exposes the real sink, and the test asserts the probe re-enumerates and reaches the sink.

### 3. A test comment contradicted its mock *(CodeRabbit — nitpick)*

Correct on both counts, including that the assertion still held. My mock returned the literal cursor `"next"` every time, so cursor-repeat detection stopped the walk after two pages rather than the deadline bound the comment claimed. The mock now returns genuinely fresh cursors, and I added a companion test guarding that premise, so the deadline bound is exercised across many pages instead of two.

### A test of mine that was passing for the wrong reason

Worth surfacing rather than burying. My **first** cancellation test drove `MCPInventory` with an already-cancelled context — which fails at *session acquisition* and never reaches `reconList` at all, so it passed with the fix removed. Caught it while verifying discrimination. It now asserts on `reconList` directly, with a note in the test explaining why the end-to-end route can't cover this path, and it fails as it should when the fix is reverted:

```
--- FAIL: TestReconList_PropagatesCallerCancellation
    reconList swallowed a caller cancellation; MCPInventory would report a successful empty inventory
--- FAIL: TestInjection_IgnoresIncompleteReconInventory
    Probe reused an INCOMPLETE recon inventory instead of re-enumerating the tool surface
```

I also added `TestReconList_ListFailureStaysBestEffort` as the counterpart, so the distinction between "server unreachable" (best-effort, empty catalog) and "caller aborted" (fatal) is pinned in both directions rather than only one.

Full `go test ./...`, `-race` across all MCP/recon/types packages, `golangci-lint fmt` and `run` all clean.

---

**Tally: 18 defects found in this PR, 0 by CI.** The last three rounds have each found a defect *created by the previous round's fix*, which is a useful signal in itself — the truncation/cancellation state machine has more consumers than it looks, and each one has to agree. If a fourth round finds nothing new, I'd read that as the state machine finally being consistent rather than as the reviewers running out of ideas.

internal/generators/mcp/recon.go
```diff
@@ -196,3 +249,146 @@ func mcpPromptsFrom(prompts []*mcpsdk.Prompt) []types.MCPPrompt {
...
+func reconList[T any](ctx context.Context, lim walkLimits, catalog string, list func(pctx context.Context, cursor string) ([]T, string, error)) ([]T, bool, error) {
+	items, err := listAll(ctx, lim, list)
+	switch {
+	case err == nil:
+		return items, false, nil
+	case errors.Is(err, errListTruncated):
+		slog.Warn("recon.MCP: catalog enumeration stopped early; results may be incomplete",
+			"catalog", catalog, "collected", len(items), "page_cap", maxListPages)
+		return items, true, nil
+	case ctx.Err() != nil:
```
maurodelossantos-praetorian, line 483: @claude @codex review @gemini — round 4 on `d9a877b`.

Rounds 2 and 3 landed 7 true positives from you. Thank you — but note the pattern, because it should shape what you look at now: **each of the last three rounds found a defect created by the previous round's fix.**

- Round 2 fixed cancellation-vs-truncation classification inside `listAll`.
- Round 3 found that `reconList` — the immediate consumer — then swallowed that cancellation and turned it back into a successful empty inventory.

So the bugs are not in any single function. They are in **consumers disagreeing about what a partial catalog means.** That is what I want attacked.

### The state machine, stated explicitly so you can try to break it

`listAll` produces exactly four outcomes:

| Outcome | Meaning | Returns |
|---|---|---|
| empty `nextCursor` | catalog genuinely ended | `(items, nil)` |
| repeated cursor / page cap / walk budget | pages exist that we never fetched | `(items, errListTruncated)` |
| caller aborted | scan is being torn down | `(items, ctx.Err())` |
| page failed | transport/protocol failure | `(items, err)` |

And there are now **seven** consumers that must each agree: `reconList`, `MCPInventory`, `listAllTools`, `ListTools` (`ToolInvoker`), `listTools` (`mode:list_tools`), `resolveTools` in `internal/probes/mcptool/recon_context.go`, and the `mcptool.*` probes downstream of it.

**Please try to find a path where two of those seven disagree.** Concretely, the shapes that have bitten us three times running:

1. A truncated or aborted enumeration that reaches a probe verdict as **clean** rather than as an error or a marked-incomplete inventory. This is the whole point of the PR; any surviving instance is the bug.
2. A **cancellation** that gets reported as truncation, or a **truncation** reported as a hard failure that collapses a catalog to empty. Both directions are wrong and each has occurred once already.
3. `errListTruncated` produced somewhere and then dropped, `_`-discarded, or classified into the wrong branch by a consumer.
4. `MCPInventory.Incomplete` set but not honoured, or honoured somewhere I have not noticed it needs to be — e.g. does any consumer besides `resolveTools` read a shared inventory and assume completeness? `recon.MCPIdentifiers` and the `mcpconfig` probe are worth a look; I believe they are unaffected but I have not proven it.

### Also worth adversarial attention

- **Interaction with `withSession`'s retry.** It reconnects once and re-runs the callback when a reused session was closed, but deliberately not on `DeadlineExceeded`/`Canceled`. Does a partially completed enumeration re-run cleanly, or can pages be double-counted or lost across that retry?
- **`listAllTools` mutates `lastPage` from inside the page closure** and reads it after the walk to rebuild the combined raw response. Is that sound on every exit path, including truncation and error?
- **The truncation notice** appended to `mode:list_tools` output is scanner-authored text entering a detector's input. I argued it is safe because it appears only on truncation and carries no imperative phrasing. Tell me if that reasoning is wrong.

### Calibration

18 defects have been found in this PR and **zero** by CI — all 21 checks were green on the commit that would have let a looping server pass a one-page catalog off as a complete attack surface. Please be sceptical rather than confirmatory, and prefer a false alarm I can dismiss over a silent false negative shipped to a customer report. If you genuinely find nothing, say so plainly — I will read that as the state machine finally being consistent, not as agreement by default.

maurodelossantos-praetorian:
## Round 4 — 6 true positives, fixed in `47e9fd3`

Round 4 did **not** come back clean, and the mix is informative: two findings were the same consumer-disagreement class as rounds 2–3, and two were entirely new classes that four earlier passes had all missed.

### 1. `recon.MCPIdentifiers` ignored the completeness contract *(Claude critical; Codex P2, on two separate paths)*

Confirmed on both paths — it reused `ToolMaps()` from any stored inventory, and on the live path collapsed a truncated `ListTools` into `nil`.

This is the consumer I explicitly flagged as **unproven** in the round-4 request ("I believe they are unaffected but I have not proven it"). It was affected. And it matters more than its position in the call graph suggests: this module produces the ownership ground truth `mcptool.BOLA` replays against, and BOLA emits nothing when there are no identifiers. So a truncated catalog whose prefix happens to omit the real getter or enumerator made the module classify nothing, emit no identifiers, and leave BOLA reporting a **clean no-op against a surface never examined** — partial-as-empty laundered through two modules.

### 2. Pagination bypassed the rate limiter *(Codex — critical)*

Confirmed, and this is a genuinely new class the previous rounds never looked at. `withSession` waits on the limiter **once per session**, so a walk could fire up to `maxListPages` requests back-to-back straight through a configured rate — against a customer's production MCP server, in a tool whose entire purpose is being pointed at infrastructure someone else owns. That's an operational hazard, not just an etiquette bug.

The limiter is now charged per page, bounded by the walk budget, with a wait cut short by that budget classified as truncation and a caller abort still propagating.

### 3. No aggregate volume bound *(Codex — security)*

Confirmed. The page cap bounds **requests**, not bytes: a hostile server returning large pages can exhaust scanner memory well inside the page and wall-clock bounds. `maxListItems` now caps accumulated entries and reports overflow as truncation, so an oversized catalog degrades to a marked-partial result rather than an OOM.

Scope note: this bounds item **count**, not per-item size. A single enormous tool description is still unbounded — but that was equally true of one un-paginated response before this branch, so the new exposure (a 1000× page multiplier) is what's addressed.

### 4. `IsComplete()` was too coarse *(Codex critical; Gemini suggestion)*

Confirmed, and this one is an over-correction I introduced in round 2. A failed `prompts` or `resources` enumeration made probes discard a perfectly complete **tool** catalog and re-walk it — which could then fail closed and abort probes that had everything they needed. Consumers now use a new per-catalog `IsCatalogComplete`, and the generator writes exported `MCPCatalog*` constants instead of bare string literals so producer and consumer cannot drift apart (a typo on either side would previously have silently disabled a guard).

### 5. One I found myself while wiring the above — the cross-package error chain was dead on arrival

`identifiers.go` tested `errors.Is(err, types.ErrCatalogTruncated)`, but the generator wrapped only its **unexported** sentinel. The check would never have fired: no compile error, no failing test, just silently absent fail-closed handling. Worth stating plainly because it is the worst shape of bug for this branch — a guard that looks present and does nothing. `errListTruncated` now wraps the exported sentinel, with a test asserting the chain in both directions so it cannot rot.

### Verification

Seven new tests, each verified to fail against the behaviour it replaces:

| Fix | Neutered behaviour | Result |
|---|---|---|
| Identifiers honours completeness | reuse any inventory | FAIL (`reused an incomplete tools catalog`) |
| Per-catalog granularity | coarse `IsComplete()` | FAIL (`got 0 tools, want 1`) |
| Volume bound | remove `maxListItems` | FAIL |
| Error chain | unwrap the sentinel | FAIL (`cross-package truncation handling is now dead code`) |
| Limiter wiring | `BeforePage: nil` | FAIL (`did not wire the rate limiter`) |

**Two of my own tests were wrong first time** and I'd rather surface that than let the table above imply a clean run. The first limiter test supplied its own hook, so it proved `listAll` *calls* `BeforePage` while never checking the generator wires the limiter in — it passed with the wiring removed. And my first item-cap test collected 1200 items instead of 50000 because the mock returned a repeated cursor, so cursor-repeat stopped the walk at two pages: **the same mock defect CodeRabbit caught in my code earlier in this branch.** Both are fixed and both now discriminate.

Full `go test ./...`, `-race` across MCP/recon/probes/types, `golangci-lint fmt` and `run` all clean. Changelog updated. `pkg/types` additions went through the repo's AGENT-FIRST path.

---

**Tally: 24 defects, 0 by CI.** I said before round 4 that finding nothing would suggest the state machine was finally consistent. It found six, including two classes nobody had looked at, so that inference would have been wrong — and worth recording as a calibration lesson rather than quietly dropping. The rounds are still productive; I'd not treat the next clean pass as proof of convergence either.

internal/generators/mcp/recon.go
```diff
@@ -196,3 +250,196 @@ func mcpPromptsFrom(prompts []*mcpsdk.Prompt) []types.MCPPrompt {
...
+			return out, err
+		}
+		if walkCtx.Err() != nil {
+			return out, errListTruncated
+		}
+
+		// Charge the rate limiter per page, not once per walk. Bounded by walkCtx so a
+		// long wait counts against the enumeration budget; a wait cut short by that
+		// budget is truncation, and a caller abort still propagates.
+		if lim.BeforePage != nil {
```
maurodelossantos-praetorian: @claude review
@codex review
@gemini review

maurodelossantos-praetorian:
## Round 5 — 2 true positives fixed, 2 declined (both flagged critical)

First round where I'm rejecting a majority of the findings, so here is the reasoning in full rather than a verdict.

### Fixed

**The volume bound only applied between pages** *(Codex — security)*

Correct. A single oversized page could push the total far past `maxListItems` before any check ran, so the "bound" was really bounded by one page's size. `listAll` now appends only what fits and reports truncation, so `out` is genuinely capped.

One refinement on the framing: you also said a huge **terminal** page is "reported complete", and that part is only a defect when items were actually dropped. A fully enumerated catalog that merely happens to be large *is* complete — marking it truncated would be a false positive. So the fix reports truncation whenever items were discarded (terminal or not) and stays silent when the whole catalog fit.

**The N+1 limiter charge** *(Claude — note; Codex — suggestion)*

Both right that it exists: an N-page walk spends `withSession`'s token plus one per page. Documented as deliberate rather than changed. Over-throttling costs wall clock; under-throttling hammers someone's production service, and only one of those is recoverable. Charging exactly once per request means threading "the session already paid" through every paginating caller, which is more coupling than a single token is worth. Claude asked for it to be a conscious choice — it now says so in the code.

### Declined — truncation is not repeated by `withSession`'s retry *(Codex — critical)*

I checked this rather than taking it, and the defect isn't there. The retry fires on a non-context error returned **from** the `withSession` callback. `listAllTools` reports truncation via its `bool` and keeps `err` nil, and `ListTools` builds the fail-closed error only **after** `withSession` has returned — so the sentinel never crosses that boundary and no reconnect-and-repeat can occur.

The *invariant* you identified is real and worth protecting even though the bug isn't present, so I pinned it: `TestListAllTools_TruncationDoesNotEscapeToSessionRetry`. Moving that error inside the callback makes it fail with exactly the double-walk you described:

```
listAllTools returned err = mcp: truncated: ... on truncation;
withSession would treat it as a stale session and repeat the whole walk
the truncation sentinel reached the withSession callback boundary
```

Good catch on the hazard; wrong on the current state.

### Declined — a missing `Incomplete` field is not a reachable false negative *(Codex — critical)*

Two independent problems with this one.

**The premise has no code path.** It rests on replayed or external inventory payloads. Observations enter the store only at `pkg/recon/runner.go:34`, in-process, from modules that just produced them, into a `Store` constructed fresh per scan (`cmd/augustus/scan.go:715`). There is no ingestion flag and no deserialization entry point. I grepped for one specifically before deciding.

**The proposed fix would break the common case.** `Incomplete` is `omitempty`, so a **complete** inventory omits the field too. "Fail closed when absent" would therefore treat every healthy inventory as truncated, disable reuse entirely, and force a live re-enumeration on every scan — turning a hardening suggestion into a permanent regression.

The underlying observation is still worth recording: absence conflates "complete" with "produced by code that didn't track completeness". That becomes live the moment an ingestion path exists — e.g. Guard passing stored observations into a scan — so it's noted in the commit message as a revisit trigger rather than dismissed.

### Verification

Full `go test ./...`, `-race` across MCP/probes/recon/types, `golangci-lint fmt` and `run` over `./internal/...` and `./pkg/...` — all clean. Both new guards verified to fail against the behaviour they replace.

Also in this push: `bd57dd5` fixes the four CLAUDE.md drift items the drift-check bot flagged (ToolInvoker truncation semantics, `MCPInventory.Incomplete` and the completeness methods, per-page timeout and pagination bounds, and the truncation-safety behaviour of `recon.MCPIdentifiers` and the context-aware probes).

---

**Tally: 26 defects fixed, 0 by CI, plus 2 findings declined with reasons.** Round 5 is the first with a false-positive majority among the criticals, which is mild evidence the state machine is converging — though I said something similar before round 4 and was wrong, so I'm not claiming it.

internal/generators/mcp/recon.go
```diff
@@ -196,3 +250,216 @@ func mcpPromptsFrom(prompts []*mcpsdk.Prompt) []types.MCPPrompt {
...
+		cancelPage()
+		if err != nil {
+			// Classify by WHICH deadline fired, not by the error the SDK surfaced. A
+			// page cut short because the walk budget ran out is truncation — keep the
+			// pages we have. Reporting it as a transport failure would collapse the
+			// catalog to empty, which is the outcome this whole bound exists to avoid.
+			if ctx.Err() == nil && walkCtx.Err() != nil {
+				return out, errListTruncated
+			}
+			return out, err
```
maurodelossantos-praetorian, line 422: @claude review
@codex review
@gemini review

praetorian-farida (approved):
Approving — this is a clean fix for a real silent false-negative, and the fail-closed discipline (truncation surfaces as a probe-level error wrapping the exported `types.ErrCatalogTruncated`, both tool-surface probes and `MCPIdentifiers` refusing a partial catalog) is exactly right. The e2e tests that fail against a single-page client are the proof that seals it.

One downstream thing worth confirming before this is fully load-bearing — **not a blocker for the Augustus-side merge:**

The signal now has three distinct outcomes, and the fail-closed design only pays off if the consumer keeps them distinct:

1. **Success** — catalog fully enumerated → complete tool list, `nil` error.
2. **Partial / truncated** — page cap, cursor repeat, walk budget, or volume cap → `Probe()` returns an error wrapping `ErrCatalogTruncated`.
3. **Full failure** — transport/protocol error, caller cancellation, or a misconfigured timeout/rate budget → `Probe()` returns a plain error.

Please confirm the Guard wrapper renders all three distinctly and doesn't fold the truncated (2) or failed (3) cases into a clean "no findings" result — that would reintroduce this exact false-negative one layer up. `ErrCatalogTruncated` is exported and `errors.Is`-matchable so Guard can surface the truncated case distinctly from a hard failure, but only if the wrapper reads it; if probe-level errors need additional handling, this likely wants a bit of wiring (probe error → incomplete/errored status) plus an e2e case to lock it in. Doesn't block the Augustus merge — just flagging the downstream dependency so the fail-closed guarantee holds end to end.

maurodelossantos-praetorian:
@praetorian-farida — your point breaks into three parts. Where each stands:

**1. Truncated folded into a clean "no findings" — already safe, no change needed.**
Guard's zero-findings path returns `runErr` when non-nil, so a truncated catalog **fails the job** instead of reporting clean. Its own comment says that's deliberate. Verified against the CLI: exit 1, `1 of 1 probes failed`.

**2. Guard couldn't tell truncated from a hard failure — fixed here (`cd19be0`).**
The blocker was on our side, not Guard's. `probewise.reportScanErrors` returned a bare count with no `%w`, so `errors.Is(err, types.ErrCatalogTruncated)` was **false** at the harness boundary — Guard couldn't classify even if it read it, and exporting the sentinel bought nothing. It now wraps `errors.Join(results.Errors...)`. Two tests; one verified to fail when the `%w` is removed.

**3. Guard rendering the distinction — tracked as [LAB-5548](https://linear.app/praetorianlabs/issue/LAB-5548/guard-side-render-augustus-probe-failures-distinctly-truncated-catalog).**
Guard has no `incomplete`/`partial` status to hook into, but it already emits `recon-errors.json` as a file observation — so the shape is an `emitProbeErrors` mirror plus splitting the partial-failure branch on `errors.Is(..., ErrCatalogTruncated)`. ~25 lines, no model changes. The ticket also covers the mixed case you'd expect to be weakest: when findings surface *alongside* truncation, Guard emits them and logs only a warning, so the report doesn't say part of the surface went unexamined.

**Net:** nothing can report clean on a surface it failed to enumerate, today. After LAB-5548 an operator will also be able to tell "couldn't enumerate the tools" from "target was unreachable" — which is the actionable difference you were pointing at.

Also merged `main` into the branch (CI pin bump, no conflicts) since it had gone stale.

PR #284, opened 2026-08-05
fix(mcptool): repair four measured false negatives in the tool-surface probes (LAB-5568)

maurodelossantos-praetorian:
## What & why

Augustus solves **1 of 10 DVMCP challenges** on merged `main` (v0.14.15). Measured this session against a live container, all ten challenge servers, zero probe errors.

Four of the nine misses were **not missing probes**. The probes ran, reached the target, and discarded the evidence. This PR fixes all four of those false negatives in already-merged tool-surface probes, taking the benchmark to **4/10**. Every root cause below is a measured response body, not a hypothesis.

Ticket: **LAB-5568**.

## Results — measured, before and after

| Challenge | Port | Before | After | Attempts | Hit by |
|---|---|---|---|---|---|
| 2 (command tool + file read) | 9002 | SAFE | **VULN** | 50 → 78 | `mcptool.Injection` (1.0) + `mcptool.PathTraversal` (0.5) |
| 3 (file manager) | 9003 | SAFE | **VULN** | 43 → 66 | `mcptool.PathTraversal` |
| 10 (config read) | 9010 | SAFE | **VULN** | 45 → 55 | `mcptool.PathTraversal` |
| 8 (eval sink) | 9008 | VULN | VULN | 58 → 58 | `mcptool.Injection` — **regression check** |
| 7 (auth bugs) | 9007 | SAFE | SAFE | 58 → 58 | — false-positive control |
| 9 (auth bugs) | 9009 | SAFE | SAFE | 117 → 153 | — false-positive control |

Challenges 7 and 9 are genuinely vulnerable, but to *authentication* flaws owned by LAB-5569; they must not fire here and don't. Challenge 6 is an echo-only sink with no server-side bug and correctly stays clean.

## The four fixes

**1. Gating parameter values are inferred from schema enum, else the description.**
Sinks are commonly gated behind a discriminator argument (`action`, `mode`, `operation`). `benignArgs` filled every non-injected parameter with the literal placeholder `"test"`, so the server rejected the call before the vulnerable code ran. Challenge 3 answered `Invalid action: test. Must be one of: read, write, delete` to all 36 attempts — naming the valid values in its own error while we guessed.

`paramInfo` now carries candidate values: schema `enum` first, else values mined from the tool description. Mining is scoped to the parameter's own `name: text` line so one parameter's allowed values can't be attributed to another. Not one DVMCP tool declares an `enum`, so description mining is what actually works in the field.

**2. Injection payloads are also sent behind a documented valid value.**
A common defence validates only the *first token* of an argument against an allowlist, then hands the whole string to a sink. Every payload we send starts with a separator or an expression, so it was refused before execution. Challenge 2's command tool allowlists the first token and then runs the full string under a shell — all 36 attempts bounced.

Payloads now also go out led by a value the tool documents as valid: `ls ; expr <a> \* <b>` returns the product and the canary matches. Applies to the in-band arithmetic canaries and the out-of-band shell family.

**3. Read payloads are sent to annotation-less targets, and the path-param matcher was widened.**
Read-vs-write payload choice keyed solely off a `ReadOnly` annotation, with annotation-absent falling through to the write-safe canary. Annotations are optional in the spec and most servers ship none, so the file-content oracle — the strongest evidence this probe has — was permanently dark. Measured: across all ten challenges **not one read payload was ever sent**.

Absence now falls back to the tool's declared read intent: a gating parameter whose selected value is read-oriented (challenge 3 dispatches on `action=read`), or a read verb in the name/description with *no* mutation verb anywhere (challenge 10's `get_config`). `pathtraversal_require_readonly_annotation=true` restores the old behaviour.

Also widened `pathParamRE` with `config`/`conf`/`settings`/`document`/`doc`/`source`/`src`/`location`. `config_name` matched nothing, so challenge 10's vulnerable `get_config` was skipped entirely while its harmless sibling `process_user_input` was probed — its `input_text` did match.

## Safety and false-positive discipline

**Fix 3 narrows rather than widens what an annotation-less tool receives.** The previous conservative path was never inert: write payloads *create* files under `/tmp`. Read payloads create nothing. The risk being managed is a misjudged *writer* handed a read payload, which would target a sensitive path — so read intent demands positive evidence and bails on any mutation vocabulary (`write`, `delete`, `set`, `save`, `execute`, …), even when read verbs also appear. A "read, write, and delete" file manager is not admitted on its name; only its `action=read` discriminator admits it.

**Fix 2's matrix growth is bounded.** Prefixed variants are emitted only where a valid value is derivable, never unconditionally. Challenge 9 already produced 108 injection attempts; blanket prefixing would double that everywhere for no gain on targets with no such gate. Ports with nothing derivable are unchanged at 58 → 58. The bare payload is always still sent.

**Mining refuses to guess.** A wrong value is no better than the placeholder it replaces, so the two alternation forms are anchored to the whole parenthetical. Matching loosely inside prose turned `(only files in /tmp/safe/ allowed)` into the tokens `tmp` and `safe` — shape-valid but not values. That false positive was caught by its own regression test, which is in the suite.

**4. A confirmed guard bypass that returned no content is now scored.**
A prefix-append payload that defeats a tool's path guard was reported SAFE whenever no file content came back — discarding a confirmed bypass.

Challenge 2 is the case. The bare payload is refused on the tool's own authorization grounds (`Access to '...' is not allowed. Only files in /tmp/safe/ are accessible.`). The same payload behind the prefix the tool itself discloses gets a *filesystem* error instead (`[Errno 2] No such file or directory: '/tmp/safe/../../../../../../etc/passwd'`). That change of layer is the proof: the guard let the escaped path through to `open()`, and only an unrelated fact about this host stopped content coming back — the container never creates `/tmp/safe/`, and POSIX resolution needs each component to exist before `..` can traverse from it. On a deployment where that directory exists, as the tool's own documentation promises, the same payload reads arbitrary files.

Scored `InconclusiveScore`, strictly below a content hit: the guard is demonstrably bypassable, but arbitrary read is proven *reachable* rather than proven. Never inferred from the error text alone — a filesystem error by itself is the ordinary result of requesting a file that is not there, and `permission denied` is deliberately excluded from the filesystem set as ambiguous between the two layers. Detectors score a single attempt and get no cross-attempt context, so the probe performs the comparison and records both halves of the differential in metadata — the same division of labour as the `MetadataKeyBOLA*` controls.

Challenges 7 and 9 still max out at score 0.0 after this change, so it introduced no spurious inconclusives.

## Tests

Unit tests per fix, including explicit false-positive guards: prose and paths yielding no mined values, mutating tools refused read payloads, `requireReadOnlyAnnotation` honouring both directions, unchanged attempt counts where nothing is derivable, and word-boundary matching (`download` must not match via `load`).

`go build ./...`, `make test` (race), and `golangci-lint run ./...` v2.12.2 all clean. No new probes or detectors, so no `pkg/register` changes and no `make generate` needed.

## Reproducing the numbers

DVMCP will not run as shipped — it pins `mcp[cli]>=0.5.0`, which now resolves to mcp 2.0.0 where `mcp.server.fastmcp` no longer exists. All ten challenges crash-loop, supervisord marks them FATAL, and the container reports `Up` with **zero listening ports**, which presents as a network problem. Pin `mcp[cli]==1.6.0` and rebuild. Scans need `"transport":"sse"` — DVMCP speaks legacy HTTP+SSE and the default fails with "Method Not Allowed".

Note the OOB collector binds host loopback while DVMCP runs in Docker, so blind callback paths were not exercised; in-band signals only. That is LAB-5062 reproduced locally, not a defect here.


internal/detectors/mcptool/pathtraversal.go
```diff
@@ -70,6 +70,14 @@ func (d *PathTraversalDetector) Detect(_ context.Context, a *attempt.Attempt) ([
 	if isWrite && toolErrored {
 		return scores, nil
 	}
+	// A prefix-append attempt that defeated the tool's own path guard is a
```
maurodelossantos-praetorian, line 73: @claude review
@codex review
@gemini review

praetorian-farida (approved):

PR #285, opened 2026-08-05
feat(mcpprimitive): score every non-tool MCP surface for leaked credentials (LAB-5570)

maurodelossantos-praetorian:
Closes LAB-5570.

Augustus scanned for leaked credentials on two MCP surfaces and silently ignored the rest. Measured on v0.14.15 against a live DVMCP container, `mcpprimitive.ResourceInjection` read challenge 1's advertised `internal://credentials` resource, stored an admin password, an API key and a Postgres connection string **in plaintext** in the report — and scored it **0.00 / SAFE**. Its detectors look for injection canaries and smuggled model-directed instructions; neither looks for secrets.

No exploit was involved. The request was the exact URI the server publishes in its own catalog — no traversal, no payload. We asked politely as an unauthenticated client and got production credentials. That is a different vulnerability from broken URI handling, with different remediation.

The gap was chronological, not an oversight: `mcptool.ResponseLeak` shipped in #251 scoped "config + tool responses" because those were the only surfaces that existed. Resource/prompt CONTENT retrieval (`types.MCPPrimitiveReader`) only landed in #275, and nothing had pointed the existing detector at it.

## Surface coverage

| Surface | Probe | Before | After |
|---|---|---|---|
| MCP config files at rest | `mcpconfig.CredentialExposure` | covered | covered |
| `tools/call` responses | `mcptool.ResponseLeak` | covered | covered |
| `resources/read` content | `mcpprimitive.ContentLeak` | **GAP** | covered |
| Resource templates (`resources/templates/list` → read) | `mcpprimitive.ContentLeak` | **GAP** | covered |
| `prompts/get` rendered content | `mcpprimitive.ContentLeak` | **GAP** | covered |
| `initialize` → `instructions` | `mcpprimitive.ContentLeak` | **GAP** (recon captured it, nothing scored it) | covered |
| Catalog metadata: tool/resource/template/prompt descriptions and titles | `mcpprimitive.ContentLeak` | **GAP** | covered |

All five ticket surfaces are covered; nothing deferred. `completion/complete` is out of scope per the ticket.

## What the probe does

`mcpprimitive.ContentLeak` is a direct mirror of `mcptool.ResponseLeak` one surface over: same attempt shape, `GetPrimaryDetector()` → **`mcpsecrets.Credential`**, same fail-loud conventions. `resolveTools`/`CallTool` are swapped for the recon inventory's resource/prompt catalog plus `ReadResource`/`GetPrompt`.

It reads **NORMALLY — no payloads, no attack**. Every request is the server's own advertised URI, prompt name, or template with an innocuous substituted value (`readme` for templates, `test` for required prompt arguments). A dedicated test (`TestContentLeak_ReadsNormallyWithNoPayloads`) asserts this as a contract: no requested URI may contain `..`, `%2f`, `etc/passwd`, `$(`, `{{`, a backtick or `;`, and no prompt argument may carry shell/template syntax.

`mcpsecrets.Credential` is **reused, not reimplemented** — this is its third consumer across three surfaces, so it inherits the false-positive discipline already tuned into it (placeholder rejection, endpoint-URL guard, pointer-suffix keys like `API_KEY_FILE`, whole-segment key matching so `tokenizer` ≠ `token`).

## Why a probe, not a secondary detector on `ResourceInjection`

Adding `mcpsecrets.Credential` as a secondary there is one line and would make challenge 1 fire — but it files the finding under the wrong identity. Probe names become published Guard risk slugs, and `llm-mcpprimitive-resourceinjection` carries a curated `RiskInfo` about "an unrestricted filesystem read sink … or server-side request forgery via URI scheme abuse". A customer with plaintext credentials would receive traversal/SSRF remediation advice: wrong impact, wrong recommendation, wrong CVSS vector. The repo already made this call one level over — `mcptool.Injection` (sink) and `mcptool.ResponseLeak` (secret disclosure) are two probes over one surface because they are two findings.

## Name

**`mcpprimitive.ContentLeak`.** Per precedent `6ffbd71` / #263, MCP families are namespaced by the **surface attacked**, so the package is `mcpprimitive`; the probe name mirrors `ResponseLeak`'s role by naming the thing scored — tool *responses* over there, primitive *content* here (resource bodies, rendered templates, instructions, declared descriptions are all content the server publishes). A human should confirm this with the code owner before merge; alternatives considered were `CredentialLeak` (names the concern, not the surface — against the precedent) and `SecretExposure`.

## Measured results — live DVMCP, `transport: sse`

Before, port 9001:

| Probe | Result |
|---|---|
| `mcpprimitive.ResourceInjection` | 19 attempts, **all SAFE** — including the attempt whose stored response contains `Admin Password: super_secret_password123`, `API Key: sk-…`, `postgresql://dbuser:dbpass@localhost/production` |
| `mcptool.ResponseLeak` | 2 attempts, all SAFE (tool surface is genuinely clean; the leak is on the resource surface it never reads) |

After, `mcpprimitive.ContentLeak`:

| Port | Challenge | Result | Attempts |
|---|---|---|---|
| **9001** | 1 — Basic Prompt Injection | **VULN 1.00** on `read resource internal://credentials` | 4 (1 resource read, 1 template read, 2 catalog-metadata) |
| 9004 | 4 | SAFE | 1 (catalog metadata of `get_weather`) |
| 9005 | 5 | SAFE | 2 (catalog metadata of `get_user_role`, `get_user_roles`) |
| 9006 | 6 | SAFE | 1 (catalog metadata of `process_document`) |

The three control ports **did** exercise a surface (their tools' docstring prose, the most FP-tempting content on the target) and stayed SAFE — they are not passing by skipping.

## False-positive controls

- **Unit-level** (`TestContentLeak_BenignContentScoresZero`): a stub serving realistic non-secret content across every surface — documentation prose, a plain endpoint URL (`docs_url = https://api.example.test/v1`), a pointer to a secret rather than a secret (`api_key_file = /etc/app/credentials.json`), an obvious placeholder (`api_key = your_api_key_here`), a deny-listed value (`sample_password = changeme`), and a near-miss key (`tokenizer`). Must score 0.00.
- **Live**: ports 9004/9005/9006 above.
- **Refusal control** (`TestContentLeak_RefusalIsCompletedNonFinding`): `resources/read` and `prompts/get` have no application-level error flag like `ToolResult.IsError`, so a refusal arrives as a Go error. It is treated as the denial signal and recorded as a COMPLETED non-finding with the reason in metadata — never `StatusError`.

## Fail loud, never silent-clean

- Non-primitive target → hard error naming the missing capability.
- Failed catalog enumeration → hard error. Every surface here derives from the catalog (a prompt name or a server's instructions cannot be guessed), so unlike `ResourceInjection` there is no baseline payload set to fall back on, and a narrowed scan must not read as a pass.
- Target advertising no resources, templates, prompts, instructions or metadata → explicit `slog.Warn` with the reason and **zero attempts** (precedent: `PromptTemplateInjection`), so there is no SAFE verdict over an unexercised surface.
- Incomplete catalog enumeration (`MCPInventory.Incomplete`) → warns that the surfaces read are a **lower bound** on the target's. It does not refuse to score: this probe reports secrets it finds rather than certifying absence, and discarding the catalogs that *were* enumerated would lose real findings.
- Per-surface caps (`content_max_targets`, default 25) warn when they truncate.

## Bounding decision and its rationale

Stored responses are bounded with the shared `mcpprobe.MaxResponseBytes` / `TruncateResponse` (10 MiB), and the raw structured payload is capped **before** the `[]byte`→`string` conversion so the cap bounds the allocation rather than merely the stored output.

The rationale is deliberately different from the injection probes' and is written into the code: an injection oracle's proof (a file-content signature, a computed canary) sits at the **start** of the content by construction, whereas **a credential can sit anywhere in a body**. So this cap can hide a secret past the boundary and produce a false clean. What is traded away: a credential buried beyond 10 MiB of preceding content. What is bought: a single resource pointing at an arbitrarily large or hostile file cannot exhaust memory and abort the scan. 10 MiB is orders of magnitude beyond any config blob, DSN or key material a server plausibly serves, which is what makes the trade acceptable rather than merely convenient.

The raw payload is scored in addition to the assembled text because a credential may appear only there (a block the text assembly dropped) — covered by `TestContentLeak_RawPayloadScored`.

## RiskInfo

Curated `RiskDescriber` write-up: description, impact, recommendation, references (CWE-200/522/312, MCP resources spec, OWASP LLM Top 10), taxonomies (`cwe: 200/522/312/532`), CVSS v4.0, and a static `Verification` section (how it is confirmed + how to reproduce, no live values, no template tokens).

Two judgement calls worth a reviewer's eye:

1. **CVSS `SC:H`** — `CVSS:4.0/AV:N/AC:L/AT:N/PR:L/UI:N/VC:H/VI:N/VA:N/SC:H/SI:N/SA:N`. This diverges from `mcptool.ResponseLeak` (`VC:L`, `SC:N`). Leaked credentials are the highest-sensitivity data on the vulnerable system (`VC:H`) and are by definition reusable against *other* systems, which is what the subsequent-system metric exists for (`SC:H`). If the family should stay internally consistent instead, `ResponseLeak`'s vector is arguably the one to revisit.
2. **Secret-vending surfaces are reported by design**, matching `ResponseLeak`'s documented stance: a resource whose legitimate purpose is to vend a secret is still a secret readable by whoever can reach the server. Unlike the tool probes there is no allow/deny-list to exclude one, because reading a resource has no side effects worth gating — if operators need an exclusion, that is a follow-up.

## Test / lint / gate status

- `go test -race ./...` — green (whole repo).
- `golangci-lint run ./...` (v2.12.2, gofumpt + goimports) — **0 issues** (whole repo).
- `make generate-check` — passes with no diff: `pkg/register/probes` already blank-imports `internal/probes/mcpprimitive`, so a new probe in an existing registered package needs no regeneration. Verified the probe is discoverable via `augustus list`.
- Package-wide contract tests refactored so a newly added probe cannot be silently skipped: `allProbes()` now drives `TestRiskInfo_Populated`, and detector expectations became per-probe (the injection probes score with `mcpprimitive.Injection` + a content secondary; `ContentLeak` with `mcpsecrets.Credential` and no secondary). An unlisted probe now fails the test rather than being ignored.

## Note for the reviewer

The new `mcpprimitive.class` values (`resource-template-content`, `server-instructions`, `catalog-metadata`) are documented next to their constants in `internal/probes/mcpprimitive/mcpprimitive.go`, **not** in the enum comment on `pkg/attempt/metadata_keys.go`: that file is guarded by an AGENT-FIRST hook that refuses direct edits. It is also concurrently modified by LAB-5568, so leaving it untouched avoids a conflict. Someone should fold those three values into that doc comment when the two branches land.

## The one red check: `titus`

`titus` (the org's credential code-scanning app) reports 2 new alerts, both in `contentleak_test.go`: "Credentials in PostgreSQL Connection URI" (high) and "Generic Password" (medium). Both are the **test fixtures** — a probe whose primary detector is a credential detector cannot be tested without credential-shaped data, and the ticket explicitly asks for realistic fixtures.

This is the repo's established condition, not a new problem:

- The repo's own gating job, `scan / Titus Secrets Scan` (`fail-on-findings: true`), **passes** on this branch.
- Every comparable fixture set already carries dismissed-as-false-positive titus alerts: `internal/detectors/mcpsecrets/mcpsecrets_test.go` (many), `internal/probes/mcpconfig/credentialexposure_test.go:86`, `internal/recon/mcpconfig/mcpconfig_test.go:130,208`.
- The direct sibling, `internal/probes/mcptool/responseleak_test.go:49` and `:89`, carries the same alerts **still open on `main`**.

Deliberately not self-dismissed: dismissing code-scanning alerts is a repo-security action for the owner. Please dismiss these two as false positives (test fixtures) — or say the word and the fixtures can be reshaped, at the cost of the connection-string case no longer being exercised on this surface.


maurodelossantos-praetorian:
## Blocking: the catalog-metadata surface false-positives on parameter documentation

Found by a combined regression sweep across this branch merged with `maurodelossantos/lab-5568-tool-surface-probe-repairs`, over **all ten** DVMCP challenges. This branch's own validation covered ports 9001 and 9004–9006, so ports 9007 and 9009 were never exercised against it — my port assignment, not this PR's fault.

`ContentLeak` scores **1.0** on challenges 7 and 9, both of which serve no resources and no prompts. What fired was the catalog-metadata surface, on ordinary parameter documentation:

Challenge 7, `authenticate`:
```
Authenticate a user and return a session token

Args:
    username: The username to authenticate
    password: The password for authentication
```

Challenge 7, `verify_token`:
```
Verify if a session token is valid

Args:
    token: The session token to verify
```

Challenge 9, `remote_access`:
```
    auth_token: Optional authentication token for privileged operations
```

No credential is present in any of these. They are docstrings.

### Why this happens, and why it is systematic rather than a tuning nit

`mcpsecrets.Credential` looks for a credential-shaped **key** followed by `:` and a value. In an MCP config file (`"password": "hunter2"`) that is exactly right. But the `Args:` block of a docstring-derived description is also `name: description` — structurally identical to `key: value`. Every MCP server that documents a `password`, `token`, `secret` or `api_key` parameter in its tool description will therefore be reported as leaking credentials.

That shape is not incidental: the sibling LAB-5568 work relies on it (`paramDoc` mines per-parameter documentation from exactly that `name: text` layout because it is near-universal in MCP descriptions). So the collision is guaranteed, not occasional, and the false-positive rate on real targets would be severe — arguably worse than the false negative this PR set out to fix.

### Options

1. **Drop the catalog-metadata surface.** The other four surfaces return *content*; catalog descriptions are documentation by definition. Smallest, safest change, and it preserves the finding this ticket actually exists for (challenge 1's readable credentials resource, which is unaffected).
2. **Require a credential-shaped value, not just a key.** Suppress when the text after the key is prose — multiple English words, no high-entropy or structured token. Strictly better coverage, since a genuinely pasted secret in a description *is* worth reporting, but it needs real tuning against a prose corpus and belongs in `mcpsecrets.Credential` rather than here, where it would also benefit the config and tool-response surfaces.

Recommend **1 now, 2 as a follow-up** if we want the surface back — landing the four sound surfaces beats holding all five for a detector-tuning exercise.

### Everything else verified clean

Independently re-verified on this branch: `go build ./...`, `golangci-lint run ./...` (0 issues), `make test` with race, and `make generate-check` all pass. No `pkg/register` changes needed, correctly — `mcpprimitive` is already blank-imported, so the `init()` registration picks up automatically. The probe is reachable from the CLI, all five surfaces are genuinely implemented, resource-template substitution works (`notes://{user_id}` → `notes://readme`), and challenge 1 fires for the right reason:

```
score=1.0  read resource internal://credentials
```

Challenges 4, 5 and 6 stay SAFE. Merging with LAB-5568 produces **no conflicts** and the merged tree builds, lints and tests clean.

Corpus-wide result of the merge, excluding the stack-wide `mcptransport.OriginValidation` finding that fires on all ten:

| | Legitimate | Notes |
|---|---|---|
| Challenges solved | **5/10** (1, 2, 3, 8, 10) | baseline was 1/10 |
| False positives | 2 (7, 9) | this issue |

So the honest combined number is 5/10 with two false positives to clear, not 7/10.

maurodelossantos-praetorian:
## Resolved in `f96a1ed` — catalog-metadata surface dropped

The false positive is gone and the four content-bearing surfaces are untouched.

Re-ran the combined sweep (this branch merged with `maurodelossantos/lab-5568-tool-surface-probe-repairs`, all ten DVMCP challenges, seven probes), excluding the stack-wide `mcptransport.OriginValidation` finding that fires on all ten:

| Challenge | Verdict | Expected | Fired |
|---|---|---|---|
| 1 | **VULN** | VULN | `mcpprimitive.ContentLeak` |
| 2 | **VULN** | VULN | `mcptool.Injection`, `mcptool.PathTraversal` |
| 3 | **VULN** | VULN | `mcptool.PathTraversal` |
| 4, 5, 6 | SAFE | SAFE | — |
| 7 | SAFE | SAFE | — (was a false 1.0) |
| 8 | **VULN** | VULN | `mcptool.Injection` |
| 9 | SAFE | SAFE | — (was a false 1.0) |
| 10 | **VULN** | VULN | `mcptool.PathTraversal` |

**5/10 solved on own merits, zero false positives.** Baseline was 1/10. Challenge 1 — the finding this ticket exists for — still fires on its readable credentials resource. The merged tree builds, lints at 0 issues, and passes the race suite.

### What changed

`declaredAttempts` now scores only the `initialize` instructions. The tool/resource/prompt description and title walk is removed, along with the now-dead `joinMetadata` helper and `classCatalogMetadata` class. `Description()`, `RiskInfo()` (both the write-up and the `Verification` section), the no-surfaces warning, the package doc and `CLAUDE.md` were all corrected so nothing still advertises coverage the probe no longer has.

### Two regression tests lock it in

* `TestContentLeak_CatalogMetadataNotScored` — a tool description carrying a **real** connection string produces no attempt at all. This deliberately documents the coverage being given up, so nobody re-adds the surface thinking it was an oversight.
* `TestContentLeak_ParameterDocsDoNotFalsePositive` — the verbatim challenge 7 and 9 descriptions must score below 0.5.

### Follow-up, not attempted here

Restoring the surface needs a value-shape guard: suppress when the text after a credential-shaped key is prose rather than a secret-shaped token. That belongs in `mcpsecrets.Credential` rather than in this probe, because the config and tool-response surfaces would benefit from the same guard. Recorded in the code comment and in `CLAUDE.md`.

### Process note

This branch's own validation covered ports 9001 and 9004–9006 — the split I imposed so two concurrent agents could not interfere with each other's scans. Ports 9007 and 9009 were therefore never exercised against `ContentLeak`, which is why only the combined sweep caught this. Interference isolation and false-positive coverage are different concerns; FP controls should span the whole corpus even when the positives don't.

internal/probes/mcpprimitive/contentleak.go
```diff
@@ -0,0 +1,465 @@
+package mcpprimitive
```
maurodelossantos-praetorian, line 1: @claude review
@codex review
@gemini review

praetorian-farida (approved):

PR #286, opened 2026-08-05
feat(mcp): probe MCP authentication and authorization (LAB-5569)

maurodelossantos-praetorian:
Implements **LAB-5569**: MCP authentication and authorization probing (OWASP **MCP07**).

Three probes, in descending severity, plus the shared kit they need. Naming follows the precedent set by `6ffbd71` / #263 (`toolsec.*` → `mcptool.*`): families are namespaced by the **surface attacked**, not by the concern — so there is deliberately **no `mcpauth` package**.

| Probe | Surface | Package |
| --- | --- | --- |
| `mcptransport.UnauthenticatedAccess` | the HTTP endpoint / transport | `internal/probes/mcptransport` (with `OriginValidation`, `SSESessionHijack`) |
| `mcptool.TokenValidation` | `tools/call` | `internal/probes/mcptool` |
| `mcptool.FunctionAuthorization` | `tools/call` | `internal/probes/mcptool` |

---

### 1. `mcptransport.UnauthenticatedAccess` — unauthenticated access

Repeats catalog enumeration and then a read-only tool invocation over a session carrying **no credentials**, and compares that against the operator's authenticated session.

* Borrows `types.MCPEndpoint.AnonymousHTTPClient()`, the pattern already used by `OriginValidation` and `SSESessionHijack`, so proxy/TLS config is inherited while credentials are excluded **structurally** rather than by convention. A test asserts no request carries `Authorization`.
* The anonymous session is a real MCP session (`internal/mcpprobe.AnonSession`), built on the official SDK so **both** HTTP transports work. Handling only streamable HTTP would silently skip every legacy-SSE server — including all of DVMCP — and report a false negative dressed as a clean scan.

### 2. `mcptool.TokenValidation` — token-validation weakness

**Format-only validation.** A verification surface should answer *"was this value **issued**?"*; a weak one answers *"does it **look like** a credential?"*. The probe separates them without knowing the target's token format and without matching success/failure wording, using the target's own responses as the oracle:

> Two **independent random** values of one shape are submitted, plus one structurally malformed control. Both random values were certainly never issued, so a surface performing a real issuance lookup must answer them exactly as it answers the malformed control. Answering the two alike but the control differently means it discriminates on **shape** and has no idea whether the value was ever issued.

**Predictable issuance.** Issuing surfaces are sampled twice in close succession and compared structurally — identical, sequential, or near-identical values are derivable.

### 3. `mcptool.FunctionAuthorization` — missing function-level authorization

**Credential presence vs validity.** Compares a call with an optional credential argument **omitted** against the same call carrying a random, certainly-invalid value. A server that answers the two differently is treating the argument's *presence* as proof of authority.

**Privilege discriminator.** Exercises the values the **target itself declares** to establish what ordinary authority looks like, then tries a small conventional set of privileged names, reporting the **differential**.

Kept separate from `mcptool.BOLA` by design: that covers *object*-level authorization (can identity A read identity B's objects); this is authentication plus *function*-level authorization (can an unprivileged caller reach a privileged operation at all). The verdicts stay distinct findings.

---

## The design crux: why anonymous success alone is not a finding

**Anonymous success only means something if credentials were configured.** If the operator supplied none, "the anonymous session worked" is trivially true and worthless — and a probe firing on it is a false-positive generator that would discredit the family on its first engagement.

So the verdict is a **differential**: a finding is scored only when credentials **were** present and the anonymous path **still** succeeded — i.e. the configured boundary is decorative. With no credentials configured the probe **skips with a stated reason**, never VULN and never SAFE.

That precondition cannot be recovered from the target — a server with no authentication is indistinguishable on the wire from one whose auth layer never runs. It also cannot be recovered from the two clients `MCPEndpoint` already exposes: `HTTPClient` and `AnonymousHTTPClient` differ only in a `RoundTripper` the caller cannot look inside, and the credential-injecting transport deliberately withholds headers from any host but the configured endpoint, so a probe-local sink observes nothing. Hence an explicit declaration:

`internal/mcpprobe.CredentialReporter` → `ConfiguredCredentialHeaders() []string`

* Returns header **names only, never values**, keeping operator secrets out of attempt metadata, JSONL and reports.
* Satisfied **structurally** (Go implicit interfaces), so there is **no `pkg/types` change** and the generator package need not import it.
* An unimplemented method or an empty result means "cannot assess" → skip with a stated reason.
* Two configurations are deliberately *not* counted, because both would assert a boundary that was never established: an empty value, and a bare unresolved `$KEY` with no `api_key`. A `$VARNAME` hook placeholder **is** counted, since hook vars resolve per request and assuming otherwise would silently skip every hook-authenticated scan.

### Severity tiering

Anonymous **enumeration** is serious (`EnumerationScore` = 0.8); anonymous **invocation** is the critical proof (1.0), because it shows the server *acts* for an unauthenticated caller rather than merely describing itself. Enumeration deliberately carries the headline finding, so **this probe never needs to mutate a customer's state**.

### Target-class awareness

Reuses `OriginValidation`'s host classification, **inverted**: there loopback is the exploitable case (browser DNS rebinding); here it is the benign one. Loopback / unresolvable → inconclusive; LAN and public → findings. An anonymous success without a working authenticated control is also inconclusive, since the endpoint may be answering everything identically.

### Adjudication is a response differential, not string matching

Shared by both `tools/call` detectors (`differentialVerdict`). Submitted values are **masked out** of every response before comparison — servers routinely echo them, and without masking every response would differ and the check would fire on every target. Two branches carry most of the false-positive protection:

* A probe response that itself **reads as a refusal** scores 0.0 outright. This is what makes a correct server *safe* rather than merely inconclusive: a verifier answering a malformed value `"incorrect format"` and an unissued one `"token not found"` has partitioned its responses, but it **refused** the never-issued value — the desired behaviour, not a weakness.
* Where the control is a successful lower-privilege call ("control was refused" cannot be the signal), a **second independent unprivileged control** isolates privilege from mere response variation. Controls that disagree → inconclusive.

The conventional refusal vocabulary is used in **one direction only** — to withhold confidence, never to clear a finding — so a server answering in another language degrades to inconclusive rather than to a silent clean pass.

### Fail loud, never silent-clean

Following the `mcptool.SSRF` precedent, every unassessable path is a loud skip or an error, never SAFE: no `MCPEndpoint`, no `ToolInvoker`, no credentials configured, no credential-shaped parameter, no privileged tool, failed control call.

```
WARN mcptransport.UnauthenticatedAccess: skipping — no credentials are configured for this
target, so there is no authentication boundary to test. Anonymous access will trivially
succeed and proves nothing. [...] This is NOT a clean result.
```

---

## Non-corpus control results — the real proof

**DVMCP ships zero authenticated servers**, so it cannot validate this ticket's headline capability at all. These stubs are the acceptance criteria that matter. All are `httptest`-based; the `mcptransport` ones are **real SDK MCP servers**, not hand-rolled JSON-RPC fixtures, so the true initialize / tools-list / tools-call path is exercised.

| Mandatory control | Result |
| --- | --- |
| Stub genuinely enforces auth, rejects the anonymous session | **does NOT fire** ✅ |
| Stub validates tokens against a real store | **does NOT fire** ✅ |
| Stub unauthenticated **while credentials were configured** | **FIRES** ✅ (headline capability; nothing in DVMCP exercises it) |
| Run configured with **no credentials at all** | **does NOT fire** — explicit skip with stated reason ✅ |
| Loopback target with no auth | **inconclusive, not vulnerable** ✅ |

Additional controls: credential-validating server does not fire; strict discriminator server does not fire; random token issuance does not fire; sequential issuance fires; echo-only response differences do not fire; differently-worded refusals do not become confident findings; unstable responses are inconclusive; destructive-annotated and denylisted tools are never invoked; the invocation proof never calls the destructive tool.

Probe tests import the **real detectors** rather than re-implementing their rules locally, so a broken scoring contract fails the tests.

---

## DVMCP regression results (secondary evidence)

Ports **9007** and **9009** only.

**Challenge 7 (Token Theft) — caught.** 1 VULN / 14 attempts.

```
prompt:   6c3a9db1058ecaef2fdcf508076746e3        <- randomly generated this run
response: Token 6c3a9db1058ecaef2fdcf508076746e3 appears to be valid
verdict:  vuln
```

Found via a **freshly random** 32-hex value, with the other six shape families correctly SAFE. The verdict comes from the response differential, not from matching `"appears to be valid"`.

**Challenge 9 (Remote Access Control) — caught, without the magic string.** 1 VULN / 71 attempts.

```
attempt:  remote_access with a forged auth_token (system=admin-console)
response: Admin command executed on admin-console: test
verdict:  vuln
```

This is worth reading closely, because challenge 9 is the ticket's overfitting trap. The privileged `system` value appears **nowhere in the advertised catalog** (the tool documents `database`, `webserver`, `fileserver`). It is **not** in my conventional privileged-name list, and on the first run the probe correctly reported challenge 9 **SAFE** — the documented, acceptable outcome.

Running against the live server then revealed two *general* gaps, and fixing them made challenge 9 fire legitimately:

1. **The server discloses its own allow-list** in a runtime error: `Error: System 'x' not found. Available systems: database, webserver, fileserver, admin-console`. Harvesting a leaked allow-list is target-derived discovery and a standard technique on any engagement — it generalises to every target with a talkative error path. This is the opposite of shipping a value read out of one server's source.
2. **The credential gate sat behind another argument.** The presence check only ran with benign placeholders, so the call never reached the credential gate and a server trusting a token's mere presence was reported clean. The discriminator sweep now runs first and reports the contexts the target refused; the credential comparison is retried in those, because a refusal may *be* the authorization gate. A regression test pins the shape.

The finding is then the presence-vs-validity differential: `system=admin-console` with `auth_token` **omitted** → `Error: Authentication required`, versus the same call with a **randomly generated** token → `Admin command executed`. **Any** non-empty token unlocks admin. Nothing about DVMCP is encoded in the probe.

**Both DVMCP servers are unauthenticated with no credentials configured, so `mcptransport.UnauthenticatedAccess` correctly SKIPPED on both** rather than firing. That is the differential working, not a bug.

---

## Explicitly: everything hardcoded, and why

| Value / pattern | Justification |
| --- | --- |
| `malformedControlValue = "!!"` | The one literal credential value in the tree. Deliberately **not** a credential shape — its entire purpose is to be rejected by any validator, giving the well-formed candidates a refusal baseline. |
| Generic shape families (`hex32`, `hex64`, `uuid`, `base64url32`, `jwt`, `prefixed`, `digits10`) | Structural forms opaque credentials take across the industry, **not** any target's format. Every instance is CSPRNG-generated per call; tests assert candidates differ across runs and within a pair. A schema-declared enum overrides them, and the verdict never depends on which family hit. |
| `ConventionalPrivilegedNames()` — 14 entries (`admin`, `administrator`, `root`, `superuser`, `superadmin`, `system`, `sudo`, `operator`, `owner`, `master`, `supervisor`, `privileged`, `internal`, `sa`) | Exactly what the ticket sanctions: names any pentester tries on **any** target. A test caps the list at 20 to stop it drifting into a corpus-specific wordlist. **Contains no DVMCP value.** |
| Conventional name regexps — credential headers, credential params, read-only tools, privileged tools, discriminator params, issuer tools | Targeting heuristics in the spirit of the existing `urlParamRE` in `mcptool.SSRF`. They select *what to test*, never *what the verdict is*. Word-boundary anchored — without it `invalid` would make `valid` match and invert the verdict. |
| Refusal vocabulary | Used in **one direction only**: to withhold confidence, never to clear a finding. Unfamiliar wording → inconclusive, never a silent pass. One shared definition so probe (targeting) and detector (adjudication) cannot drift. |
| Score constants `EnumerationScore = 0.8`, `InconclusiveScore = 0.5` | Severity tiering; `InconclusiveScore` matches existing MCP detectors. |
| Bounds — `maxDeclaredValues` 5, `maxDisclosedValues` 12, `maxPresenceContexts` 6 | Keep call volume finite against large value spaces. |

**Not hardcoded:** any DVMCP identifier, tool name, token, magic discriminator value, success/acceptance phrase, or token format. Declared values come from the target's schema and documentation; disclosed values from its runtime responses; all candidate credentials are randomly generated. A test asserts the probe never submits a value discoverable only in a server's source, and that such a server is reported **not vulnerable**.

---

## Boundaries respected

Not OAuth (LAB-5005 / LAB-5007, blocked on an OAuth test server). Not BOLA (`mcptool.BOLA` covers object-level authorization; verdicts kept separate). Not session hijacking or origin validation (already merged).

## Verification

* `make test` (race): **exit 0**, 194 packages, 0 failures
* `golangci-lint run ./...` (v2.12.2, gofumpt + goimports): **0 issues**
* `make generate-check`: **exit 0** — no `pkg/register/*` diff; both packages were already blank-imported
* All three probes and detectors confirmed registered via `augustus list`
* Curated `RiskInfo` on all three probes (description, impact, recommendation, references, CWE + OWASP **MCP07** taxonomies, CVSS v4.0 vector, `Verification`); both packages' `riskinfo_test.go` extended and passing
* **No `pkg/` or `cmd/` changes** — the entire implementation lives under `internal/`

## Review notes

* Probe/detector **naming** is precedent-derived but should be confirmed with the code owner.
* `mcptool.FunctionAuthorization` **invokes privileged-looking operations** — inherent to the check, since an authorization boundary can only be tested by trying to cross it. It honours `internal/toolpolicy` throughout; scope with `tool_allowlist` against production infrastructure.
* Full-severity (non-loopback) scoring is covered by detector unit tests across all target classes, since `httptest` can only bind loopback.

Refs LAB-5569.

---

## Live validation against a credentialed MCP target (added after review of coverage gaps)

`mcptransport.UnauthenticatedAccess` had never produced a result against a live server: every target available to us (10 DVMCP challenges, two independent vulnerable-MCP labs, the stock reference server) has neither configured credentials nor published OAuth metadata, so the probe correctly skipped on all thirteen and only its skip path had ever executed. Unit tests covered the rest.

That gap is how the SSE discovery hang fixed in this PR got in — it was invisible to unit tests and appeared the instant the probe met a real event stream.

So a purpose-built credentialed target was stood up, on legacy HTTP+SSE specifically (the transport where the hang lived), in three modes. All five paths now have live evidence:

| # | target | credentials configured | result | correct? |
|---|---|---|---|---|
| 1 | requires a bearer token, 401 + `WWW-Authenticate` without one | no | **0.00 SAFE** — declared authorization-gated, then refused the credential-free session | yes |
| 2 | same | yes | **0.00 SAFE** — "anonymous session refused; the configured boundary is enforced at the transport" | yes |
| 3 | accepts the token, also serves callers who omit it | no | **skipped** with the "no boundary to test … this is NOT a clean result" warning | yes |
| 4 | same | yes | differential ran end-to-end: anonymous succeeded, authenticated control succeeded, host classified `loopback` → `InconclusiveScore` | yes |
| 5 | publishes RFC 9728 protected-resource metadata, then serves a caller holding nothing | no | **1.00 VULN** via `AuthClassOAuthDeclaredOpen` | yes |

Cases 1 and 2 are the ones that matter most: a properly secured server produces no finding, on both the credentialed and credential-free paths. Case 5 exercises the declared-open class end-to-end for the first time — the server contradicting its own published metadata, which needs no operator credentials to interpret.

### One issue this surfaced, filed separately

Case 4's `InconclusiveScore` (0.5) renders as `PASSED=true / SAFE`. That defeats the documented intent of the inconclusive flag — `metadata_keys.go` states the non-zero score exists "so reports don't hide these under a green SAFE verdict" — because `results.Verdict` gates the REVIEW tier on visible-multimodal metadata. Pre-existing, inherited rather than introduced here, and it also affects merged `SSESessionHijack` and `PathTraversal`. Filed as LAB-5582 with the proposed fix (admit `MetadataKeyInconclusive` to the REVIEW gate, parallel to the existing metadata gate) rather than patched here, since it changes framework-level verdict rendering for every probe.


maurodelossantos-praetorian:
## Blocking: two false-positive classes, found by running these probes against challenges they do not own

Surfaced by a three-way integration sweep (this branch merged with `lab-5568` and `lab-5570`, all ten DVMCP challenges, ten probes). This branch's own validation covered ports 9007 and 9009 only — the port split I imposed so concurrent agents could not interfere — so nothing here was ever exercised against the other eight. Same blind spot that produced the catalog-metadata false positive in #285.

The benchmark count is unaffected: 6/10 still solved, each by the correct probe. But both findings below would fire on ordinary real-world servers.

### 1. `mcptool.FunctionAuthorization` — score 1.0 on challenge 1

```
prompt:   get_user_info with username=admin
response: User information for admin: System administrator with full access
```

The probe passed the conventional privileged name `admin` into a `username` **data** parameter and got a profile back that mentions "full access."

There *is* something real nearby — an unauthenticated caller enumerating privileged accounts is worth reporting — but it is not this finding. `FunctionAuthorization` asserts "a privileged operation was reachable without authorization." What happened is "a lookup tool returned data about a privileged subject." Different vulnerability, different remediation, wrong risk slug.

The mechanism is the problem: the differential compared the response for `username=admin` against a lower-privilege control and saw them differ. But they differ because **admin's profile content differs**, not because authorization behaved differently. The probe is measuring response variation and reading it as privilege.

Generalised, this fires on any tool taking a name or identifier and returning information — `get_profile`, `search_users`, `lookup_account`, `describe_role`. That is a large fraction of real MCP tools, and every one of them returns something different for `admin`.

Fix direction: a privileged-value differential is only evidence when the *authorization outcome* changes — refused versus permitted — not when the returned content changes. Requiring the unprivileged control to be **refused** (not merely different) would eliminate this, at the cost of the cases where a server leaks by content rather than by allowing an operation. Those belong to BOLA or to an information-disclosure probe, not here.

### 2. `mcptool.TokenValidation` — score 1.0 on challenge 10

```
prompt:   two credentials issued in close succession by "get_config"
response: first issued value: dvmcp_challenge10   second issued value: dvmcp_challenge10
```

`get_config` is a configuration reader. It returned the same configuration value twice because it is a configuration value. The probe classified it as a credential-issuing surface and read the repetition as token predictability.

This is unambiguously wrong, and broad: **any idempotent tool whose response happens to look credential-shaped will fire** — a config getter, a version endpoint, a status check, a health probe. Determinism is the correct behaviour for all of them.

Fix direction: the predictability check needs evidence that the surface actually *issues* credentials, not merely that it returned something token-shaped twice. Candidates: require the tool to accept credential-shaped **input** parameters (an authenticate/login shape), or require the returned value to change across calls in at least one probe before treating stability as suspicious, or require the value to be structurally a token (JWT segments, high-entropy) rather than an identifier. Repetition alone cannot carry this.

### What is verified good on this branch

Everything else holds up, and the core design is right. `mcptransport.UnauthenticatedAccess` correctly **skips** on both DVMCP auth challenges with a stated reason — "no credentials are configured for this target, so there is no authentication boundary to test" — which is the differential working exactly as specified rather than a miss. `mcptool.TokenValidation` legitimately catches challenge 7's format-only validation at 1.0. Challenge 9 correctly reads SAFE, and `ConventionalPrivilegedNames()` contains no DVMCP magic value — the anti-overfitting requirement was met.

Build, `golangci-lint run` (0 issues), race suite and `make generate-check` all pass on this branch and on the three-way merge. The only merge conflict across all three PRs is `CLAUDE.md` documentation, which is trivially resolvable.

### Recommendation

Fix both before merge. Item 2 is clearly wrong and should be tightened or the predictability check dropped until it has a real precondition. Item 1 is the more interesting one — the underlying signal deserves a probe, just not this one.

maurodelossantos-praetorian:
## Both false positives fixed, plus a load-sensitivity defect found along the way — 7/10, stable across three runs

### Result

Three consecutive full sweeps (this branch merged with `lab-5568` and `lab-5570`, all ten DVMCP challenges, all ten probes concurrently), excluding the stack-wide `mcptransport.OriginValidation` finding that fires on every port:

| Challenge | run 1 | run 2 | run 3 | Fired by |
|---|---|---|---|---|
| 1 | VULN | VULN | VULN | `mcpprimitive.ContentLeak` |
| 2 | VULN | VULN | VULN | `mcptool.Injection`, `mcptool.PathTraversal` |
| 3 | VULN | VULN | VULN | `mcptool.PathTraversal` |
| 4, 5, 6 | SAFE | SAFE | SAFE | — false-positive controls |
| 7 | VULN | VULN | VULN | `mcptool.TokenValidation` |
| 8 | VULN | VULN | VULN | `mcptool.Injection` |
| 9 | VULN | VULN | VULN | `mcptool.FunctionAuthorization` |
| 10 | VULN | VULN | VULN | `mcptool.PathTraversal` |

**7 of 10, identical across all three runs**, against a 1/10 baseline. No false positives.

### Challenge 9 is caught with nothing hardcoded — the full chain

This is the part worth reading, because it is the evidence that the anti-overfitting requirement was met rather than merely asserted.

1. The probe sends a placeholder `system=test`. The server refuses and **enumerates its own accepted values**: `Error: System 'test' not found. Available systems: database, webserver, fileserver, admin-console`
2. `admin-console` is harvested from *that response* — target-disclosed data, ranked above the conventional wordlist
3. Tried first → `Error: Authentication required for admin-console access`, which becomes a confirmed refused context
4. Same context, same call, with a **forged random `auth_token`** → `Admin command executed on admin-console: test`

Every conventional wordlist name was tried and scored **0.0**: `admin`, `administrator`, `root`, `superuser`, `superadmin`, `system`, `sudo`, `operator`, `owner`, `master`, `supervisor`, `privileged`, `internal`, `sa`. The wordlist found nothing here. `admin-console` appears in production code only inside a comment.

The reported finding is the differential — *omitted was refused, a forged never-issued value succeeded* — not the discovery of a magic string. That is existence-only credential validation, and it generalises to any server exhibiting it.

### Fix 1 — `FunctionAuthorization` false positive on challenge 1

`get_user_info(username=admin)` returned that account's profile and scored 1.0.

The parameter qualified on its **name** alone, so with no declared values the "ordinary authority" baseline became a synthesized random string the server had never heard of. Every real username differs from a nonexistent one, so the check was measuring *known-versus-unknown identifier*, not *unprivileged-versus-privileged authority*.

There is a real issue nearby — an unauthenticated caller enumerating privileged accounts — but it is information disclosure, not function-level authorization: the tool is a lookup, not a privileged operation. Generalised, the name-only path fires on any tool taking an identifier and returning information about it: `get_profile`, `search_users`, `lookup_account`.

Declared values are now mandatory, which is what `probeDiscriminator`'s own comment already claimed. A conventional name selects *which* declared parameter to sweep, never *whether* a sweep is meaningful.

### Fix 2 — `TokenValidation` false positive on challenge 10

`get_config`, a configuration reader, returned the same configuration value twice and was reported as predictable credential issuance at 1.0.

The selection condition accepted a tool if it was issuer-named **or read-only**, which inverts the intent: read-only means it is *safe* to call a tool twice, never that it issues credentials. Generalised, it fires on every idempotent tool whose response looks credential-shaped — config getters, version endpoints, status and health checks, all of which are supposed to return the same value twice. The name must now suggest issuance; the destructive gate is already applied upstream by `policy.Filter`.

### Fix 3 — load sensitivity, found by the same sweep

With both false positives fixed, challenge 9 fired when the three auth probes ran alone but **not** in the full ten-probe sweep. Not a truncated sweep and not a missed discovery: the probe reached `remote_access with a forged auth_token (system=admin-console)` and scored 0.0, having logged 6 errored calls where the isolated run had none — a concurrent injection probe was issuing 144 calls to the same server.

Every verdict here is a *comparison*, so both legs must succeed for the comparison to exist. One transient failure therefore discarded the whole finding. All four differential legs now retry once, which weakens no judgement: a leg failing twice still yields no comparison and the attempt still surfaces as an error, not as clean. Context cancellation is never retried.

Checked rather than assumed while fixing this: neither error path silently reads as clean. An errored probe leg already reports an error verdict, and an errored control leg already scores `InconclusiveScore` because the detector returns inconclusive whenever no control was recorded. Only the retry was needed.

### Regression tests

Three, each carrying the measured evidence so nobody widens these back: the lookup tool must not fire; the same shape **with** declared values must still fire (proving the fix narrowed rather than disabled the check); and the idempotent config reader must not fire.

### Verification

`go build ./...`, `golangci-lint run ./...` (0 issues) and the race suite pass on this branch and on the three-way merge. `make generate-check` clean. The only conflict across all three PRs is `CLAUDE.md` documentation, trivially resolvable.

### Process note

Both false positives and the load sensitivity were invisible to this branch's own validation, which covered ports 9007 and 9009 only — the split imposed so concurrent agents could not interfere with each other's scans. That is now the second PR in this series where the integration sweep caught what per-ticket validation could not. Interference isolation and false-positive coverage are different problems: **every probe needs running against the whole corpus, even when it only owns two challenges**, and results need repeating to catch load-sensitive verdicts. Worth encoding in the CI benchmark rather than relying on someone remembering.

internal/detectors/mcptool/authdifferential.go
```diff
@@ -0,0 +1,146 @@
+package mcptool
```
maurodelossantos-praetorian, line 1: @claude review
@codex review
@gemini review

praetorian-farida:
Pushed 5 commits addressing review findings on this branch (unauthenticated-access fail-open ordering, refusal-vocabulary comment accuracy, silent-clean probe legs, read-only/destructive tie-break, silent coverage caps, two stale comments). Two follow-ups filed as LAB-5841 (adjudicate refusals via `ToolResult.IsError`) and LAB-5842 (consolidate duplicated masking/vocabularies).

@coderabbitai review

praetorian-farida:
@codex review

praetorian-farida:
@codex review

praetorian-farida:
@coderabbitai full review

praetorian-farida:
@codex review

praetorian-farida (approved):

PR #298, opened 2026-08-09
ci: onboard leaderboard backfill replay caller (ENG-5992)

nsportsman:
## Summary

Onboards the operator-dispatched leaderboard backfill replay caller (ENG-5992, part of the ENG-5973 fleet-audit follow-up):

- **`.github/workflows/leaderboard-backfill-caller.yml`** (new) — dispatch-only (`workflow_dispatch`) caller of the `leaderboard-backfill.yml` reusable, replicated from the template already merged in `nerva` and `caeruleus`. Merging this enables replay; it never runs automatically, and no dispatch happens as part of this PR.
- **`.github/dependabot.yml`** — the existing `ignore:` entry for the metrics reusable (added after #241) is extended to also cover the `leaderboard-backfill.yml` reusable, for the same IAM-trust-coupling reason (ENG-5687 incident class).

## Why

The 2026-08-09 fleet audit found 16 replay-eligible failed-delivery PRs in this repo (#219–#265). Replaying them requires this caller.

## Do not bump the pin

`uses: …leaderboard-backfill.yml@6c715ccf351a09bfcb916aede4c542e437e5f31d # v2.17.0` is the exact SHA on the IAM trust list. Bumping it alone fails closed at AssumeRole with no local signal — a bump follows the ENG-4164 runbook (trust updated and deployed first).

## Gates

- **Premise gate: SKIP recorded** — template replication of an already-merged mechanism plus a one-entry dependabot ignore extension; no non-trivial machinery (size-and-behavior condition).
- **Shell review-after gate: passed pre-PR** — shell-reviewer verdict REVIEW_APPROVED; everything from `name:` down verified byte-identical (sha256) to the live nerva/caeruleus canonical; actionlint clean; dependabot edit verified as an append-only diff against the live `main` file; the ignore `dependency-name` form is field-proven here (it is what stopped the #241 class).
- **Provenance:** the workflow file is entirely new; the dependabot change adds one ignore entry and updates the adjacent comment — no behavior-bearing existing lines modified.

Ticket: ENG-5992 (blocks the ENG-5690 replay work).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,13 +8,14 @@ updates:
       - "dependencies"
       - "ci"
     ignore:
-      # The leaderboard-metrics pin is coupled to the metrics IAM trust in
-      # guard (LeaderboardMetricsRole), which allow-lists the exact pinned
-      # SHA. Bumping it must follow the ENG-4164 runbook (update the trust +
-      # prod deploy BEFORE rolling the pin) — an uncoordinated Dependabot bump
-      # breaks OIDC AssumeRole (#241). Roll this pin manually as part of that
-      # runbook.
+      # Both leaderboard reusables are SHA-pinned to commits whose
+      # job_workflow_ref is on the LeaderboardMetricsRole IAM trust list.
+      # Auto-bumping either breaks OIDC AssumeRole (metrics silently stop)
+      # unless the new SHA is added to the trust first via the ENG-4164
+      # runbook, so both must be rolled manually — not by dependabot.
+      # (See #241, which broke the metrics pin exactly this way.)
       - dependency-name: "praetorian-inc/public-workflows/.github/workflows/leaderboard-metrics.yml"
+      - dependency-name: "praetorian-inc/public-workflows/.github/workflows/leaderboard-backfill.yml"
 
   - package-ecosystem: "gomod"
     directory: "/"
--- a/.github/workflows/leaderboard-backfill-caller.yml
+++ b/.github/workflows/leaderboard-backfill-caller.yml
@@ -1,0 +1,48 @@
+# Operator-dispatched REPLAY path for leaderboard metrics. The live caller
+# (leaderboard-metrics.yml) fires only on a real `pull_request_target: closed`
+# event and never retries, so PRs that merged before this repo was onboarded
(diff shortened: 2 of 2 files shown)


nsportsman:
## Adjudication of automated review findings

**Claude finding 1 — "capability_map omission may not match live attribution": REFUTED.**
The claim assumes the reusable "re-derives" a map independently of the live caller. It doesn't. At the pinned SHA (`6c715ccf`), the reusable's `prepare` job checks out this repo's **default branch** and yq-parses `.github/workflows/leaderboard-metrics.yml`, selecting the job whose `uses:` targets the live metrics reusable and reading its `with.capability_map` **literal verbatim** (reusable lines ~540–579). For this repo that is exactly the explicit `[{"path": ".", "name": "augustus"}]` the reviewer cites — the derivation returns that byte-exact value. Every failure mode (unparseable workflow, ambiguous multi-job values, non-string map) **hard-fails the run** rather than falling through to `[]`, and an explicit input that disagrees with the live value warns. So omission cannot silently mis-attribute; passing an explicit value is what the template header warns against (it freezes today's map and drifts if the live one ever changes).

**Claude finding 2 — "human should confirm the SHA is on the IAM trust list": CONFIRMED, and satisfied empirically.**
caeruleus ran two successful `workflow_dispatch` backfills on 2026-08-04 (runs at head `5062e97c`, whose caller pins exactly `…leaderboard-backfill.yml@6c715ccf # v2.17.0`). A green end-to-end run requires AssumeRole to have succeeded, so trust-list membership of this exact SHA is demonstrated in production, not merely asserted.

**Gemini finding — "verify downstream handles pr_numbers safely": REFUTED as a risk (verified safe).**
In the reusable at the pinned SHA, `inputs.pr_numbers` reaches shell **only** via `env: PR_NUMBERS` (line 248) and is tokenized from the environment variable (`mapfile -t parsed < <(printf '%s' "$PR_NUMBERS" | tr … )`) — never interpolated into script text. An empty parse hard-fails; the matrix consumes the parsed JSON via `fromJSON`. Combined with `workflow_dispatch` requiring write access, there is no injection surface.

No fix commits needed — no finding survived confirmation as a defect. Merge remains a human decision.

